PR #1, opened 2016-06-09
Extend meta-ivi-test with a test that generates CAPI code during build

pkonopel:
The last two patches (4f0454f and 3c833bd) can be applied on top of c55c45f independently.  The native recipes allow using Common API C++ generators as a part of the tool chain on the build host.  They rely on pre-built generators being available through the CAPIC++ update site.

The recipe capicxx-perf builds two executables (client and server) that rely on CAPIC++ with the D-Bus backend.  It invokes the code generators during the build and can be used to [smoke-] test both the CAPIC++ runtime and tools for core and D-Bus.

The recipe capicxx-someip-native can be optionally added to support SOME/IP code generator.  It is not required by capicxx-perf, though.

### DIFF
--- a/meta-ivi-test/recipes-extended/common-api/capicxx-core-native_3.1.5p2.bb
+++ b/meta-ivi-test/recipes-extended/common-api/capicxx-core-native_3.1.5p2.bb
@@ -1,0 +1,10 @@
+SUMMARY = "Common API C++ core generator"
+
+LAUNCHER_BASE = "commonapi-generator"
+LAUNCHER_LINK = "capicxx-core-gen"
+
+require capicxx-native.inc
+
+SRC_URI = "${BASE_URL}/generator/${BASE_VER}/${PV}/commonapi-generator.zip"
+SRC_URI[md5sum] = "c12551e8f61da944b17a22c00d3e6559"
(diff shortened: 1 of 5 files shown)


tolkien-joh:
I checked capicxx-perf is working.
one thing I ask you.

could you move all test binary from /usr/bin to /opt/test/... like below?
--- a/meta-ivi-test/recipes-extended/common-api/capicxx-perf_git.bb
+++ b/meta-ivi-test/recipes-extended/common-api/capicxx-perf_git.bb
@@ -14,3 +14,21 @@ DEPENDS = "common-api-c++ common-api-c++-dbus capicxx-core-na
 inherit cmake pkgconfig

 EXTRA_OECMAKE += "-DCMAKE_INSTALL_PREFIX=/usr"

FILES_${PN} = " \
    /opt/tests/${PN}/\* \
    "

FILES_${PN}-dbg += " \
    /opt/tests/${PN}/.debug/\* \
    "

do_install_append() {
    _DEST=${D}/opt/tests/${PN}
    _BIN=${WORKDIR}/build
    install -d ${_DEST}
    mv ${D}/usr/bin/\* ${D}/opt/tests/${PN}
    rmdir ${D}/usr/bin
    rmdir ${D}/usr
}

pkonopel:
OK, the PR is updated as requested and rebased on the current master (40f4544).

PR #2, opened 2016-06-10
64-bit and multilib build

cledra:
The purpose of this set of patches is to fix some installation errors which occur when using multilib. When doing so, the traditional '/usr/lib' installation directory may be replaced by variants such as '/usr/lib32' or '/usr/lib64'. Thus, hard-coded '/usr/lib' in Makefiles, recipes or config files may lead to build errors. With this patches we have tried to generalize the use of Yocto's environment variables for setting the library installation directory.

Along the same line, some references to the ${PN} variable were changed to ${BPN}, as ${PN} may be suffixed with things like -native, -cross, or prefixed with lib32-, leading to broken URLs.

We remain open to any suggestion as for how the same result could be achieved in a cleaner way. For example, another way to do commit a606699 would be to just define LIB_SUFFIX in the *.bb file:
    EXTRA_OECMAKE += "-DLIB_SUFFIX=${@d.getVar('baselib', True).replace('lib', '')}"
So we don't need to patch CMakeLists.txt.

Tests:

With these modifications we could successfully build the QEMU image with the following configurations:
- MACHINE=qemux86
- MACHINE=qemux86-64
- MACHINE=qemux86-64 with multilib for 32-bit applications

For this last build we added the following lines in conf/local.conf:
require conf/multilib.conf
MULTILIBS = "multilib:lib32"
DEFAULTTUNE_virtclass-multilib-lib32 = "x86"


gunnarx:
This is a great contribution, but adding .patch files to meta-ivi to fix various build issues doesn't feel that nice to me - that should be more of an emergency solution IMHO.   It's better to get the component fixed.

An important first step is to notify each upstream mailing list about this proposal.  Ideally each project is maintaining their own recipe inside meta-ivi, and they might have some suggestions for you.  Or preferrably they will take the patch directly:

```
{genivi-ipc,genivi-diagnostic-log-and-trace,genivi-lifecycle,genivi-ivi-layer-management}@lists.genivi.org 
```

For some of the parts (keyutils) I didn't find any other location of the recipe.  (It is not in meta-openembedded at least?).   So I suppose the change make sense - I'm not sure if anyone is communicating with the keyutils project about OpenEmbedded/Yocto.

${BPN} I have no opinion - if the change is needed then +1.

gunnarx:
cledra - thanks for notifying the mailing lists and for clarifying the purpose of ${BPN} in the email.  Now we wait a little for (hopefully) comments and feedback from each component owner.

tolkien-joh:
Hi
to verify this patch, I have to prepare test build for qemux86, qemux86-64.
it takes more time to confirm it.

tolkien-joh:
there is build failure
 when I try to build MACHINE=qemux86-64 with multilib for 32-bit applications

ERROR: QA Issue: audiomanager: Files/directories were installed but not shipped in any package:
  /usr/lib
  /usr/lib/pkgconfig
  /usr/lib/audiomanager
  /usr/lib/pkgconfig/audiomanager.pc
  /usr/lib/audiomanager/cmake
  /usr/lib/audiomanager/cmake/audiomanagerConfig.cmake
Please set FILES such that these items are packaged. Alternatively if they are unneeded, avoid installing them or delete them within do_install.
audiomanager: 6 installed and not shipped files. [installed-vs-shipped]

cledra:
Indeed, the same problem occurs with audiomanager.
I would fix this problem by adding a patch in meta-ivi, which would modify CMakeLists.txt to make it use "${CMAKE_INSTALL_LIBDIR}" (see patch in attachment).
Note that this solution only works because Poky defines ${CMAKE_INSTALL_LIBDIR} for us in cmake.bbclass. 
I understand that it would be better to fix CMakeLists.txt upstream. However, outside of a Yocto/Poky environment, "${CMAKE_INSTALL_LIBDIR}" might not be defined, so I don't know what is the clean way to do this.
[0001-audiomanager-fix-lib-install-path-for-multilib.patch.zip](https://github.com/GENIVI/meta-ivi/files/323280/0001-audiomanager-fix-lib-install-path-for-multilib.patch.zip)

Anyway, I asked the genivi-audio-manager mailing list for their opinion.

tolkien-joh:
Hi Clement Dransart

about the patch of audio manager, It may not accepted to audiomanager guys.

in meta-ivi, it's ok because we are based on poky.
could you rebase these patch and upload it again?

cledra:
Hi @tolkien-joh ,
I uploaded the patch.

tolkien-joh:
Hi @cledra 

there is an issue of gtest build
 when "MACHINE=qemux86-64 with multilib for 32-bit applications" configuration is used.
therefore, I cannot get test binary of wayland-ivi-extension that is based on gtest.

other builds are ok.

anyway, because meta-ivi does not support "MACHINE=qemux86-64 with multilib for 32-bit applications",
 I ignore this issue at this time.

tolkien-joh:
it's a workaround what I used. (it ignores the error only. not fix it)

[0001-gtest-workaroud-for-MACHINE-qemux86-64-with-multilib.zip](https://github.com/GENIVI/meta-ivi/files/329609/0001-gtest-workaroud-for-MACHINE-qemux86-64-with-multilib.zip)

tolkien-joh:
@cledra 

would you like to rebase your patch and uploaded again without any modification?
then, I'll merge it.

PR #3, opened 2016-06-15
Merge 10.0 into master

gunnarx:
The purpose of this pull request is to test automated PR build check in go.genivi.org for meta-ivi.
(But updating master to match 10.0 is also a good idea, so feel free to merge it.)


meta-ivi/recipes-support-ivi/ecryptfs-utils/ecryptfs-utils_106.bb
```diff
@@ -43,3 +43,5 @@ do_install_append() {
         install -m 0644 ${WORKDIR}/ecryptfs.service ${D}${systemd_unitdir}/system
     fi
 }
+
+RDEPENDS_${PN} = "nspr nss"
```
chbae, line 47: It's better to add in DEPENDS.

PR #9, opened 2016-07-18
Make the setting of EXTRA_USERS_PARAMS a default value

RobertAJMarshall:
[GDP-287] Enables the setting of an empty root password for testimage
as well as allowing the user to build images with other root passwords


tom--pollard:
This is both useful for the testimage CI viewpoint and for the normal users (i.e set a custom root password at buildtime via local conf explicit set). I cannot see anything within the git history that should prevent this change.

tolkien-joh:
I confirmed.

if I added EXTRA_USERS_PARAMS = "" to conf/local.conf,
 there is no root password.

with nothing to do so, root password is still existed.

PR #12, opened 2016-07-27
10.0

tolkien-joh:
hot fix


tom--pollard:
+1 for the hotfixes to the recipe branches, node-startup-controller also have the branch issue in 11.0

tom--pollard:
I've found that removing ;branch=$ in the src url and adding ";nobranch=1" fixes the issue of not finding the sha in the listed branch. I propose this is an easier and smaller changes (requires less testing of new commits)

PR #13, opened 2016-07-27
hot fix of 9.0.1

tolkien-joh:


tom--pollard:
I've found that removing ;branch=$ in the src url and adding ";nobranch=1" fixes the issue of not finding the sha in the listed branch. I propose this is an easier and smaller changes (requires less testing of new commits)

PR #18, opened 2016-08-22
Capicxx native fix

tolkien-joh:


gunnarx:
Issue added for the component here:
https://at.projects.genivi.org/jira/projects/COM/issues/COM-11
If upstream can't fix it it in the short term I guess we have to live with this hack for a little while...

slawr:
+1 for merge

tolkien-joh:
merge it now.
if there is an update for this issue on common-api-generator until Sept 17, 2016.
I'll revert it if possible.

tom--pollard:
I'm finding this patch commit unbuildable on 14.04 (within GDP) is there anything extra needed? Currently have openjdk 7 on the host

tom--pollard:
(backporting 1.8 allows the install to pass).... If the java version was not 1.8 I'd expect the statement to pass, so I'm unsure where the exit 1 is coming from

RobertAJMarshall:
I think the java version test should read
   java_version=`java -version 2>&1 | head -n 1 | grep { '[ "]1\.8[\. "$$]' || true; }`
so that bitbake doesn't get upset when it spots that grep is returning 1

slawr:
@tom--pollard "unbuildable" is not a very descriptive issue report :)

gunnarx:
Documenting for the future:

This was not changed for the previous versions ,but from 3.1.11 released now the software uses Java-8.  
I am guessing it *only* support Java-8 now but that is not clear.  So be aware, when/if the upgrade is made to 3.1.11:

Relevant info here:
https://at.projects.genivi.org/jira/browse/COM-11  Support for Java 8

PR #20, opened 2016-09-21
Fix build issue with java versions other than 1.8

RobertAJMarshall:
[GDP-398] bitbake was spotting that grep was returning 1
This commit avoids that issue

### DIFF
--- a/meta-ivi/recipes-extended/common-api/capicxx-native.inc
+++ b/meta-ivi/recipes-extended/common-api/capicxx-native.inc
@@ -26,7 +26,7 @@ DD = "${D}${datadir}/${PN}-${PV}"
 
 do_install() {
     # work around for java-8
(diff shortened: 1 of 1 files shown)


tom--pollard:
This change allows me to build the latest meta-ivi11 on ubuntu14.04

tolkien-joh:
} cause build error like below:

.../x86_64-linux/capicxx-dbus-native/3.1.5p2-r0/temp/run.do_install.21377:
     command substitution: line 106: syntax error near unexpected token `}'

RobertAJMarshall:
I'm sorry - does

java_version=`java -version 2>&1 | head -n 1 | grep '[ "]1\.8[\. "$$]' || true;`

then work for you? I'm not sure why a bitbake capicxx-dbus-native -c install
wasn't giving an error for me with that line in my commit (note that there's back quotes around the command, github formatting...)

tolkien-joh:
@RobertAJMarshall 
it works.

PR #21, opened 2016-11-09
Delete 0001-net-bus-add-the-AF_BUS-socket-address-family.patch

gunnarx:
AF_BUS is no longer part of the platform.  No reason to keep the patch in the file tree.

Signed-off-by: Gunnar Andersson <gand@acm.org>


gunnarx:
Sorry, screwed up the pull request.  I will resubmit.

PR #22, opened 2016-11-10
Delete AF_BUS patches and commented lines

gunnarx:
AF_BUS is no longer part of the platform. No reason to keep the patches in the file tree.
I separated the changes to D-Bus and kernel into different commits.


tolkien-joh:
there is no impact on build s/w.

PR #23, opened 2016-11-25
Remove baseline_ci_helper and include checkout script in meta-ivi/scripts instead

gunnarx:
meta-ivi does not include any "init.sh" like GDP and it also does not use submodules like GDP.  Instead there are manual steps described in the README.

To make the build automatic, we must extract the right commit hash from the README, for poky and for meta-openembedded.

I have previously created https://github.com/genivi/baseline_ci_helper to be able to automate the build in https://go.genivi.org.  

1. ...but those scripts seem useful to all users of meta-ivi.  
2. Also, meta-ivi already has a scripts/ directory we can use.
3. Also, it is better to not have too many repositories on GitHub. (@jeremiah)  So after this we could maybe remove the baseline_ci_helper from GitHub.

Therefore I propose to include this in meta-ivi.

I also have an init.sh script that will execute these checkouts.  I can add this if you want - then it will be very similar to GDP - only one setup script necessary.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,28 @@
+# Scripts
+
+## checkout_layer_hash.sh
+
+A build helper for meta-ivi.
+
+This script extract the revision information for dependent layers (poky
+and meta-openembedded) from the README of meta-ivi so that automated builds
+are possible.
+
+Usage:
+
+Run in meta-ivi repository root (meta-ivi, not meta-ivi/meta-ivi).
+The script assumes you have cloned poky and meta-openembedded inside
+meta-ivi root which is normally not the case in a full system build - 
+then meta-ivi is in parallel with all other layers, but when building
+the baseline only, we assume the build happens inside of meta-ivi dir.
+
+
+Example:
+$ cd meta-ivi 
+   (the first one)
+$ checkout_layer_hash.sh poky
+$ checkout_layer_hash.sh meta-openembedded
+
+## Other scripts
+
+* See README.md in project root
--- a/scripts/checkout_layer_hash.sh
+++ b/scripts/checkout_layer_hash.sh
@@ -1,0 +1,107 @@
+#!/bin/sh
+# (C) Gunnar Andersson 2015-12-12
+# License: MPLv2
+
+# This fetches information about layer dependencies
+# as documented in meta-ivi README.
+# revision, branch name, or sub layer name can be extracted
+
+README=README.md
+DEBUG=false
+
+
+fail() {
+   echo "*** FAILED ***"
+   [ -n "$1" ] && echo "Message: " $@
+   exit 1
+}
+
+sanity_check_num_lines() {
+   what=$1 ; min=$2 ; max=$3
+
+   numlines="$(wc -l)" # < Reads from stdin
+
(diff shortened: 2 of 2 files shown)


tolkien-joh:
Hi @gunnarx 

this is not working on.
it may work with following condition:
 1. this script can be run in the directory that have README.md of meta-ivi
    that means meta-ivi/ directory.
 2. poky/ and meta-openembedded/ directory should be below meta-ivi/ directory
 3. user has already gotten poky/ and meta-openembedded/ from repository
 4. current build system of meta-ivi/ does not work with above directory layout.
    meta-ivi/ -- meta-ivi/
                  +- meta-ivi-bsp/
                  +- poky/
                  +- meta-openembedded/
                  +- build/

gunnarx:
Yes, I agree.  Those are the conditions.  The Go.CD build scripts set up everything like this so it is all I need.

What do you want to do for other users?
1. Create a more complete script which will clone poky and meta-openembedded, if they do not exist? 
or 
2. Document how to use it in the README?
or 
3. Something else?

gunnarx:
Actually, the README I provided already explains the conditions, does it not?

tolkien-joh:
I see. then, in that context,
 I want for you to be clear that this script could be run on go-cd server only.

gunnarx:
I see the problem now.  I could change the script so that it works correctly with:

/dir/meta-ivi/
/dir/poky
/dir/meta-openembedded
(Go server setup)

But the script should also work with:

/dir/meta-ivi/
/dir/meta-ivi/poky
/dir/meta-ivi/meta-openembedded
(Is the second example the preferred setup for baseline?)

I could change the script so it works with both setups.

gunnarx:
I don't know, maybe my README is also describing it wrong?

OK, please clarify which directory structure you prefer for normal baseline build and I will redo this to make the script support it.

tolkien-joh:
@gunnarx 

I checked the script again.
to run this script in my development environment,
 I have to modify a little.

could you check it again?

```
diff --git a/scripts/checkout_layer_hash.sh b/scripts/checkout_layer_hash.sh
index bafba31..596fa73 100755
--- a/scripts/checkout_layer_hash.sh
+++ b/scripts/checkout_layer_hash.sh
@@ -30,7 +30,7 @@ sanity_check_num_lines() {
    fi
 }
 
-[ -f ./$README ] || fail "Cannot find README.md.  (Current directory is $PWD)"
+[ -f $README ] || fail "Cannot find README.md.  (Current directory is $PWD)"
 
 get_layer_info() {
    layer=$1
@@ -102,6 +102,6 @@ revision=$(get_layer_info $checkoutlayer revision)
 # poky to be found inside the root of meta-ivi, that is the $PWD when
 # this script is run.  (Not true on full system builds, but typically
 # when building baseline only)
-cd "$checkoutlayer" || fail "Can't cd to layer directory ($checkoutlayer)"
+cd ../"$checkoutlayer" || fail "Can't cd to layer directory ($checkoutlayer)"
 git checkout $revision
 
```

tolkien-joh:
it is merged at 12.0 branch.

https://github.com/GENIVI/meta-ivi/commit/8a3ffa132bdc4cb482c58888871dd8291479164b

PR #32, opened 2017-04-09
poky-ivi-systemd.conf: enable aarch64 for qemu-native

phongt:
The QEMU_TARGET is lacking of aarch64 that gobject-introspection
need for g-ir-scanner-qemuwrapper.

[GDP-585] Pipeline fail for Renesas Gen3 starter kit
          build (needs qemu-aarch64?)

Signed-off-by: Phong Tran <tranmanphong@gmail.com>

### DIFF
--- a/meta-ivi/conf/distro/poky-ivi-systemd.conf
+++ b/meta-ivi/conf/distro/poky-ivi-systemd.conf
@@ -36,7 +36,7 @@ DISTRO_EXTRA_RDEPENDS_append_vexpressa9 = " ${POKYQEMUDEPS}"
 
 TCLIBCAPPEND = ""
 
-QEMU_TARGETS ?= "arm i386 x86_64"
+QEMU_TARGETS ?= "arm aarch64 i386 x86_64"
 
 XSERVER ?= "xserver-xorg \
            xserver-xf86-config \


gunnarx:
Thanks @phongt, looks like exactly what we need.  We should test this with GDP next.
+1 from me.  @tolkien-joh, if you see no issues here then a quick merge would be appreciated.

tolkien-joh:
Hi all.

it's ok for build. I've got such patch last Wednesday.

but, does qem-system-aarch64 support graphics display?

gunnarx:
> but, does qemu-system-aarch64 support graphics display?

I do not know, but it might at least overcome a build problem we had.  Thanks.

gunnarx:
@tolkien-joh Could you cherry-pick this to 11.0 branch also?  It seems like it would be just as useful there and it would also make GDP 11 supported.

tolkien-joh:
@gunnarx 
I did.

see https://github.com/GENIVI/meta-ivi/commit/f293f122237a3aeb3a5a48812d4a23488d76a0e7

gunnarx:
oh, you are right.  Sorry I guess I forgot to fetch the latest.

slawr:
Thank you for merging this @tolkien-joh. 

You asked a sensible question about gfx support. I think we have two separate but related cases here. In this specific case qemu-aarch64 is required at build time for gobject-introspection. See GDP-585 for details. Then there is the general case of adding qemu aarch64 as one of the meta-ivi build targets as we discussed recently in the BIT and there the question of gfx support is important.

I think we can treat this as two separate steps. In GDP-585 we were wondering about host dependencies, as Yocto is usually good at building what it requires. On some systems there was a build failure when the distro did not provide qemu-aarch64. So if this commit solves that then it is very useful.

PR #34, opened 2017-04-18
persistence-client-library: Update to latest commits on GitHub

gunnarx:
* Update references after migration of upstream source from git.projects.genivi.org to GitHub
* Build the PCL tools (by specifying the --enable-tools option to configure)

Signed-off-by: Gianpaolo Macario <gianpaolo_macario@mentor.com>


gunnarx:
General comment on build checks:  As we have seen before, if "nostromo-image" is attempted the build of course fails because this PR is for the 11.0 branch which has build instructions only for "miranda-image".  The image changes name with each branch.  Clearly a build-helper script or some other simple logic in the CI system would improve this but that's room for improvement.  For now we have separate pipelines for Miranda and Nostromo, one of which should succeed in the normal case.

Miranda-Image is till failing however due to SDL issue on Ubuntu based build machine.
That's fixed on 12.0 branch: https://github.com/GENIVI/meta-ivi/commit/298629e1262e1e5ba1461e664c1af70c26bd8a9d but needs to be on 11.0 also.  If that is done, I'll rebase this PR.

PR #35, opened 2017-04-22
The connectivity check URI needs to change

gunnarx:
The CONNECTIVITY_CHECK_URIS variable was set to check bugs.genivi.org 
but that site is now decomissioned.  A new alternative URL is needed because now it
seems that bitbake stops working right away.

Signed-off-by: Gunnar Andersson <gandersson@genivi.org>


gunnarx:
Of course this fix is needed also on other branches 10.0, 11.0, master ...

gunnarx:
I am merging the fix since this is preventing all builds and is therefore urgent.

PR #39, opened 2017-05-30
persistence-client-library: fix QA errors

GordanM:
QA Issue: persistence-client-library: Files/directories were installed but not shipped in any package:
  /Data
  /Data/mnt-c
  /Data/mnt-wt
Please set FILES such that these items are packaged.
Alternatively if they are unneeded, avoid installing them or delete them within do_install.


tolkien-joh:
that files are include in ${PN}-test packages.

see https://github.com/GENIVI/meta-ivi/blob/master/meta-ivi-test/recipes-extended/persistence-client-library/persistence-client-library_%25.bbappend

GordanM:
@tolkien-joh thanks for pointing that out!

PR #43, opened 2017-06-29
ivi-logging: Adds initial ivi-logging version

frznlogic:

### DIFF
--- a/meta-ivi/recipes-support-ivi/ivi-logging/ivi-logging.bb
+++ b/meta-ivi/recipes-support-ivi/ivi-logging/ivi-logging.bb
@@ -1,0 +1,34 @@
+#
+#   Copyright (C) 2015 Pelagicore AB
+#
+#   SPDX-License-Identifier: MIT
+#
+DESCRIPTION = "IVI-Logging is a software which facilitates the generation \
+of log data. It provides an easy-to-use C++ API enabling you to send \
+log data to various channels such as to the console, files, or GENIVI's \
+DLT"
+HOMEPAGE = "https://github.com/Pelagicore/ivi-logging"
+LICENSE = "MPL-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=815ca599c9df247a0c7f619bab123dad"
+
+PR = "r0"
+SRC_URI = "${GIT_REPO};branch=master"
+SRCREV = "7ec74df57744189bdccb77a184beceefaf446ba3"
+PV = "0.1+git${SRCREV}"
+
+DEPENDS = "glib-2.0"
+
+inherit cmake
+
+# DLT support is optional
+PACKAGECONFIG ??= "dlt"
+PACKAGECONFIG[dlt] = "-DENABLE_DLT_BACKEND=ON,,dlt-daemon,"
+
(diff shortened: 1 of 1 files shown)


meta-ivi/recipes-support-ivi/ivi-logging/ivi-logging.inc
```diff
@@ -0,0 +1,23 @@
...
+PR = "r0"
+
+DEPENDS = "glib-2.0"
+
+inherit cmake
+
+# Make DLT support optional
+# To enable, create a .bbappend with PACKAGECONFIG_append = "dlt"
+# in your project layer
+PACKAGECONFIG ??= ""
```
GordanM: In meta-ivi context I would include the `PACKAGECONFIG ?= "dlt"` and remove the above comment or edit it so the user can exclude `dlt` if wanted. In the GENIVI layer (meta-ivi) I guess it should be on by default.

GordanM (commented):
Nice @frznlogic , could you open an issue in https://github.com/Pelagicore/meta-bistro that the recipe was copied here and has to be removed from meta-bistro to avoid duplicates.

tolkien-joh:
Hi @frznlogic 

is this what implementation of compliance of GENIVI?
could you give me a link to GENIVI compliance spec?

thank you.

meta-ivi/recipes-support-ivi/ivi-logging/ivi-logging.inc
gunnarx: Agreed, dlt should be default.  Introducing ivi-logging was a GDP decision however.  I guess we should see which components in meta-ivi will be affected before the recipe is required here.  There are probably some, but let's take it in the right order.

gunnarx:
I provided a pointer to Tolkien about the specification.  Let's discuss if and when this component should go into meta-ivi.  In my opinion it should be added when there is another meta-ivi component that depends on it.

meta-ivi/recipes-support-ivi/ivi-logging/ivi-logging.inc
```diff
@@ -0,0 +1,23 @@
+DESCRIPTION = "ivi-logging"
+HOMEPAGE = "http://www.pelagicore.com"
```
gunnarx: DESCRIPTION ought to be a bit more verbose. 1-2 sentences.

HOMEPAGE should point to a proper software project home page for this component, otherwise it does not serve any purpose IMHO.  It would be better to leave it out.  If you think it's important then I suppose you could put the home page link in the comment header, although it's not customary in bitbake recipes to have a lot of extra info in comments.

chbae:
Why do you separate include and bb file? I recommend that please use just bb file.

frznlogic:
Updated the commit according to the comments on them on a technical level. Regarding the compliance spec etc, see gunnarx comment. 

There might be a point to adding this somewhere in a higher layer for now, but I do believe it belongs somewhere down the line such as meta-ivi or meta-ivi-common.

frznlogic:
Closing this as it's moved to https://github.com/GENIVI/genivi-dev-platform/pull/104 in genivi-dev-platform instead

PR #45, opened 2017-07-03
cairo: fix QA Issue

GordanM:
Delete files from removed packages. Otherwise bitbake will terminate with
the following error:

cairo do_package: QA Issue: cairo: Files/directories were
installed but not shipped in any package:
  /usr/lib/cairo

### DIFF
--- a/meta-ivi/recipes-graphics/cairo/cairo_%.bbappend
+++ b/meta-ivi/recipes-graphics/cairo/cairo_%.bbappend
@@ -9,8 +9,10 @@ PACKAGES_remove =+ "\
 do_install_append () {
 	rm -f ${D}${bindir}/cairo-trace
 	rm -f ${D}${libdir}/cairo/libcairo-trace.so*
-
-	rmdir ${D}${bindir}
+        rm -f ${D}${libdir}/cairo/*.la
+ 
+        rmdir ${D}${bindir}
+        rmdir ${D}${libdir}/cairo
 
 	rm -f ${D}${libdir}/libcairo-script-interpreter.so*
 }


GordanM:
Discussion from openembedded: http://lists.openembedded.org/pipermail/openembedded-core/2017-July/139049.html

tolkien-joh:
@GordanM 

with this patch, I've got the following error message:

"
rmdir: failed to remove ‘/storage/src/CCCP/genivi-baseline_12/build/tmp/work/core2-64-poky-linux/cairo/1.14.6-r1/image/usr/lib/cairo’: Directory not empty
"

GordanM:
@tolkien-joh I've tried reproducing your issue and was not successful. Did you clean the sstate explicitly before trying to bitbake the recipe with the patch applied?

tolkien-joh:
@GordanM sure.

tolkien-joh:
sorry. this patch is already applied in other PR[#54](https://github.com/GENIVI/meta-ivi/pull/54)

tolkien-joh:
sorry. this patch is already applied in other PR#54

GordanM:
@tolkien-joh , I honestly disagree with the technical solution with the merged PR. But as I am not the maintainer I do not have a mandate to decide which solution will be merged. As you have previously seen in the mailing-list thread between Ross and me (found in one of the comments above), the folder should be deleted only if empty and otherwise kept in tact. Removing the folder completely will delete the .so files that are packaged in case of dev builds. However I might be wrong. 

Though, I do believe that a discussion should be made somewhere, preferably as a part of this PR, instead of opening new PRs or even more so merging them without a discussion. The community should able to see why one solution is better than the other and why was the PR created previously not fixed but a newly staged one was accepted. The process of creating new PR requests and merging them for the sole reason of a passing build makes me question the process and my further involvement in contributing to this component. 

Adding @jeremiah and @gunnarx to the thread as you might find this relevant.

slawr:
Hi Gordan (@GordanM)

As the Genivi Baseline Integration Team (BIT) Lead I know that Tolkien has just started holiday, till the 16th, so wanted to let you know we will not get any input from him for a while and also to provide some initial response to the points you raise.

Firstly, I want to say it is good to see people using meta-ivi and as importantly putting changes back to improve it. So thank you from the BIT for doing just that.

I recall cairo related issues in the past, but I should say I have not delved deeply into this specific issue beyond a first quick read of the PRs and the Yocto ML thread.

I hear two main concerns in your most recent response:
1. There is a better way to fix the issue than the one that was merged.
2. Concerns about the process.

Taking the first, I would say the code has not been set into concrete. If it can be improved further then we can do that. All that is required is a proposal. Whether a change is achieved by reverting Oleksandr's commit in favour of yours, or by simply improving the code as it is now is academic (process and how we got here aside) as either route requires changes. As Tolkien is now on holiday my personal suggestion would be to raise a JIRA issue - that concentrates discussion in one place rather than over multiple PRs - and spend the time discussing how to improve it.

Am I right in thinking Oleksandr and yourself both work for the same company?
Could I suggest you both get together and try to come to a conclusion on the best way forward?

Reading the ML thread I wonder why the recipe is not more robust about the removal of a package from the image, but that is a minor point. Balanced against other things that could be improved I would think ignore that, fix the issue and move on :)

Your PR is against the master branch which means Genivi 12. I wonder why Tolkien or I had not seen this issue. Any idea on why it would not show up all the time?

Regarding the process, I can only make some general comments as we don't have Tolkien to provide his thinking. Although I think I understand the gist of your comments I have a harder time following some of it. There does seem to have been some discussion, although not perhaps to the "ok which do we use?" poll you would like. Personally I like to see JIRA tickets for issues. It provides a single place for discussion of the problem, which can be difficult when there are multiple PRs. It also typically provides more detail about the problem itself. The commit message only provides scant details and relies on reading the ML thread (not referenced in the commit msg) to understand some of the nuances.

Although you kindly commented on #54 that a related PR already existed I don't see any review comments from you as to why you thought your PR was an improvement. I'm guessing you felt that was implicit by inference from the ML thread. I think explicit review comments from yourself would have flagged that more clearly.

Yes, the reasoning for choosing one over the other could have been more explicit. Personally I file that under "do better next time".

Tolkien is not here and I have no way of knowing what conversation you may have had by email. I would say that his comment "no build error with this patch on my desktop PC" on #54 was made on various other open PRs. We previously had a problem on the CI whereby the PR pipeline was failing due to a missing gzip dependency with the dlt-daemon build. I thought that had been fixed but seems to have come back in some form and thus all the automated build checks for all PR's have been failing with that same - unrelated to the PR - issue. So I believe in posting that to all the PRs he merged in the last couple of days he is simply saying "the CI check fails, but it builds for me locally". I don't read that as PR #54 builds, PR #45 fails so I take #54.

On the wider general process I would say we deliberately try to be pragmatically flexible. I've worked in Genivi long enough to see frustrating times where we have had more formally rigorous processes only to see them ignored or not get the input we would like. Indeed the new president has stated he wants imperfect contribution over less contribution. Maybe that sometimes creates short falls on either side.

Worse case the BIT meets every Weds at 10am CET. I am always happy to schedule a discussion about a tech point such as this if you want to discuss.

Ultimately I hope you don't allow a single issue to dampen your enthusiasm for the project.

Regards

Steve
BIT Lead

jeremiah:
It seems to me, and please correct me if I'm wrong, that the two patches are roughly similar. Tolkien may have had an issue with one because #45 doesn't use the -f flag (force) while #54 does.

#45 adds;
```
+        rm -f ${D}${libdir}/cairo/*.la
+        rmdir ${D}${libdir}/cairo
```
While #54 adds;
`rm -rf ${D}${libdir}/cairo`

rm -rf is nasty :-) but it works. rmdir ${D}${foo} feels more yocto-ish and safer and I'd prefer it but sometimes expediency is the better part of valor. Or something. 

Sorry to be milquetoast about all this -- why don't we just add the -f flag to #45, revert #54 and call it a day?

PR #46, opened 2017-07-03
dlt-daemon: remove inherit gzipnative

GordanM:
The class was renamed to gzip-native or pigz-native, which is already a
dependency. Using later verisons of gzip, found in poky 2.3, do not have a
class called gzipnative. By removing it we ensure an error free build.

### DIFF
--- a/meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.16.0.bb
+++ b/meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.16.0.bb
@@ -21,8 +21,7 @@ SRC_URI = "git://github.com/GENIVI/${BPN}.git;protocol=https \
     "
 S = "${WORKDIR}/git"
 
-inherit autotools gettext cmake systemd
-inherit gzipnative compress_doc
+inherit autotools gettext cmake systemd compress_doc
 
 PACKAGES += "${PN}-systemd"
 SYSTEMD_PACKAGES = "${PN} ${PN}-systemd"


gunnarx:
There is a build failure related to *gzip* in the configure stage.  The build should not depend on the host having gzip installed I suppose?  I have the feeling I recognize this from earlier times when the same fix was created.  We have looked at this before, but I don't know if it was solved.  Somehow the pigz-native dependency does not make this work but I do not know why.  Any idea?

tolkien-joh:
In my build environment,
 with or without this patch, I build dlt-daemon without error.

GordanM:
@tolkien-joh and @gunnarx it seems quite confusing that in Poky from Morty to Pyro version they have decided to remove the gzipnative.bbclass (http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/gzipnative.bbclass?h=morty) and make sure that `gzip` exists even if the native recipe is stated as a dependency. 

In morty the bbclass just enforces the population of the tool which seems quite weird.

@tolkien-joh building meta-ivi with this commit against Poky Morty will have issues so I would suggest that this commit be included in a later release that is targeting Poky Pyro, as the commit message explains the bbclass doesn't exist anymore.

Sorry for the confusion but it seems like the upstream project made it a bit difficult for others.

gunnarx:
@tolkien-joh wrote:
> In my build environment, with or without this patch, I build dlt-daemon without error.

So I guess the recipe is finding gzip in your build host?  I don't think it is enough to say "it is OK in my environment" without finding out the reason.  It's important to find out what the true cause of a problem is before a change is merged because sometimes the CI system is the *correct* one.   We have had cases before where the build *should* fail, because it was depending on a host tool that was not the minimal dependency for Yocto. This happened on another recipe when ImageMagick was used from the host!  That's not a tool we can assume is installed and it definitely not a good idea to have host tools leaking into the build.  

Yocto is improving the repeatability and avoiding host leakage better in pyro and later, but if it's enough I don't know yet.

For this case I don't understand it because the gzip binary seems to be installed on the build agents as expected, but maybe it is looking for something like a library, or headers, I don't know.

The build agents are anyway based on a minimal image that installs only the tools that are [required by the Yocto manual](http://www.yoctoproject.org/docs/current/mega-manual/mega-manual.html#packages)

Unfortunately, the Yocto manual seems not fully consistent.  For Fedora, it says to install gzip.  For Ubuntu/Debian it does not say that gzip needs to be installed, which is surprising.  Therefore gzip installation is not part of the tools for the build agents.  Different distros could also package things differently and include  -devel packages or not.

I think it would be better if all of the dependencies were listed clearly for all distros, and will ask Yocto project about this.

gunnarx:
> it seems quite confusing that in Poky from Morty to Pyro version they have decided to remove the gzipnative.bbclass (http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/gzipnative.bbclass?h=morty) and make sure that gzip exists even if the native recipe is stated as a dependency.

@GordanM Agreed. What's even more confusing is that the [mailing list discussion](https://lists.genivi.org/mailman/private/genivi-meta-ivi/2017-March/001422.html) linked from  [DLT-4](https://at.projects.genivi.org/jira/browse/DLT-4) states that gzipnative was removed already in Morty, but the ticket ends with a patch that *adds* gzipnative to the dependencies!

Since things change from branch to branch I have a problem with this, and I've pointed this out before, that it's not clear how *master* branch is used by meta-ivi.  Since you sent the PR to master, which Yocto branch is it supposed to be compatible with, Morty or Pyro?   

Is it possible to please decide and document what master is, in relation to the named branches that are under development?  <-- @slawr for info.

gunnarx:
Just confirming that for the [most recent build](https://go.genivi.org/go/pipelines/yocto-baseline-GITHUB-PR-Check/6/Build/2) of this, on go-agent1, the gzip binary exists in /bin/gzip on the host, but that does not help this build issue, so I guess it is something else...

GordanM:
> Since things change from branch to branch I have a problem with this, and I've pointed this out before, that it's not clear how master branch is used by meta-ivi. Since you sent the PR to master, which Yocto branch is it supposed to be compatible with, Morty or Pyro?

@gunnarx I have to admit I find this quite confusing as well. In the commit message it is stated that this commit is relevant for builds against Poky 2.3 (Pyro). Would it be a good idea to create meta-ivi branches that follow the Yocto Project/OpenEmbedded branches and stage the release branches/tags separate from that? I can understand that the GENIVI release process is different and it cannot follow the Yocto Project, but having compatible branches would be really nice. As an example: If I would want to include meta-ivi in a legacy build against Poky Krogoth what version should I use?

PR #44, opened 2017-07-03
node-state-manager: Depends on gdbus-codegen

GordanM:
The node-state-manager build will fail if glib-2.0-native is not stated
as an explicit dependency because the gdbus-codegen tool is not available.

| gdbus-codegen --interface-prefix=org.genivi.NodeStateManager.
--c-namespace=NodeState --generate-c-code=generated/NodeStateConsumer
--generate-docbook=doc/NodeStateConsumer
model/org.genivi.NodeStateManager.Consumer.xml
| gdbus-codegen --interface-prefix=org.genivi.NodeStateManager.
--c-namespace=NodeState
--generate-c-code=generated/NodeStateLifecycleControl
--generate-docbook=doc/LifecycleControl
model/org.genivi.NodeStateManager.LifecycleControl.xml
| gdbus-codegen --interface-prefix=org.genivi.NodeStateManager.
--c-namespace=NodeState
--generate-c-code=generated/NodeStateLifecycleConsumer
--generate-docbook=doc/LifecycleConsumer
model/org.genivi.NodeStateManager.LifecycleConsumer.xml
| /bin/bash: gdbus-codegen: command not found

Signed-off-by: Gordan Markuš <gordan.markus@pelagicore.com>

PR #47, opened 2017-07-09
node-state-manager: add dependancy to glib-2.0

sashko:
Compilation requires gdbus-codegen, which is in the glib-2.0 package.


GordanM:
Already merged https://github.com/GENIVI/meta-ivi/pull/44.

PR #50, opened 2017-07-29
gstreamer: add glib-2.0-native to depends

sashko:
Package requires glib-genmarshal, the lack of which causes the following
build error:

  /bin/bash: line 1: glib-mkenums: command not found
  Makefile:1884: recipe for target 'gstenumtypes.c' failed


sashko:
Failed due to the issue in a different package, dlt-daemon. Could you please restart the build?

tolkien-joh:
there is no issue with this patch on my desktop PC.

PR #51, opened 2017-07-29
gstreamer1.0-plugins-base: add glib-2.0-native to depends

sashko:
Package requires glib-genmarshal, the lack of which causes the following
build error:

  /bin/bash: line 1: glib-mkenums: command not found
  Makefile:1270: recipe for target 'video-enumtypes.c' failed


tolkien-joh:
there is no build failure with this patch on my desktop PC.

PR #53, opened 2017-07-31
dlt-daemon: remove obsolete compress_doc inheritance

sashko:


tolkien-joh:
no build fail with this patch on my desktop PC

PR #54, opened 2017-07-31
cairo: fix package installation

sashko:
The bbappend causes the following issue on do_install:

cairo do_package: QA Issue: cairo: Files/directories were installed but
not shipped in any package:

/usr/lib/cairo


GordanM:
@sashko there already exists a commit that address this issue and it's in review: https://github.com/GENIVI/meta-ivi/pull/45 .

tolkien-joh:
no build error with this patch on my desktop PC

PR #56, opened 2017-08-03
dlt-daemon: disable man pages generation

sashko:
Otherwise, it still produces the following error:

| CMake Error at doc/CMakeLists.txt:78 (MESSAGE):
|   Could not find gzip for man page compression.

### DIFF
--- a/meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.16.0.bb
+++ b/meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.16.0.bb
@@ -35,7 +35,7 @@ SYSTEMD_SERVICE_${PN}-systemd = "dlt-example-user.service \
     dlt-receive.service"
 SYSTEMD_AUTO_ENABLE_${PN}-systemd = "disable"
 
-EXTRA_OECMAKE = "-DWITH_SYSTEMD=ON"
+EXTRA_OECMAKE = "-DWITH_SYSTEMD=ON -DWITH_MAN=OFF"
 
 FILES_${PN}-doc += "/usr/share/dlt-filetransfer"
 


tolkien-joh:
@sashko 

please, confirm that https://github.com/GENIVI/meta-ivi/commit/4c1871b3f2da59f0a2927f855134d647877bed71 is not effective or not

and, show your build environment.

sashko:
@tolkien-joh if you'll take a look at the [CMakeLists.txt](https://github.com/GENIVI/dlt-daemon/blob/master/doc/CMakeLists.txt#L69), you find the following if code:

```
if(WITH_MAN)
	# Compress the man pages and install to proper place
	FIND_PROGRAM(GZIP_TOOL
	             NAMES gzip
	             PATHS /bin
	                   /usr/bin
	                   /usr/local/bin)
```
Therefore if we wanna get rid of this compilation error, setting WITH_MAN to off would be the best way to do it.

gunnarx:
I have also seen this error a lot lately in a CI environment.  I could not at first understand the problem because the CI environment even has /usr/bin/gzip installed.  But as far as I am aware,  yocto now builds in a more proper sandbox so that host tools are not available.  It could be that gzip needs to be added to the [HOSTTOOLS variable](http://www.yoctoproject.org/docs/current/mega-manual/mega-manual.html#var-HOSTTOOLS)...

Since meta-ivi is being built for the embedded target I would still propose that generating man pages is unnecessary.  It is wrong to do it if we want a minimal system...   So I would vote for  -DWITH_MAN=OFF being added back in any case.

gunnarx:
Fixed go.genivi.org pipeline to include meta-gplv2 and build-check will run again.

jeremiah:
FWIW, I've done some packaging of DLT for debian for GENIVI members who're interested in evaluating GENIVI software for their debian based system. Debian throws an lintian warning when there is an executable without a man page. Since this is a development platform and since "embedded" has changed somewhat over the years, I'd prefer to have all binaries come with man pages.

gunnarx:
> FWIW, I've done some packaging of DLT for debian for GENIVI members who're interested in evaluating GENIVI software for their debian based system. Debian throws an lintian warning when there is an executable without a man page.

That makes perfect sense for Debian which is still _primarily_ used on desktop.  What is required in Debian packaging does not affect which build flags to give to Yocto.

You making available Debian packages with man-pages make it easier to get the information, which is good, if they are not on the target system.

> Since this is a development platform and since "embedded" has changed somewhat over the years, I'd prefer to have all binaries come with man pages.

Changed in terms of power, size, disk space sure.  But embedded has not changed in the sense that the final embedded systems should include only what they need.  You would not ship man pages on an embedded system.

With reference to "development system", the only question that remains is if people are reading man pages on the target itself.  I do not think they do.  If GDP were a self-hosted system in which you actually do the majority of the development work on the system itself, that mightbe a different story, IMHO.

gunnarx:
I'd rather provide all man pages in a host development environment but if they are going on GDP, then _all_ man pages should really be there.

Right now, meta-ivi need to figure out how to solve the build issues.  Please make it work consistently or disable the man generation...

@tolkien-joh, I don't think https://github.com/GENIVI/meta-ivi/commit/4c1871b3f2da59f0a2927f855134d647877bed71 works when compiling with pyro?

sashko:
@tolkien-joh it's been 3 weeks; can we assist you somehow? we need those patches! thanks.

tolkien:
@sashko I don't want to drop manual pages of it. is there any option except "turn off" manual?

tolkien:
the reason why there are different result of dlt-daemon build may be found by comparing the builder's development environment. then, we could know this patch is good or not.

to do so, we create https://at.projects.genivi.org/jira/browse/BASE-50
@sashko please, describe your build PC's environment on that page.

tolkien:
Hi @sashko 

I met this issue during release of meta-ivi 12.0.1, and
fix it. review https://github.com/GENIVI/meta-ivi/commit/4a64b490832e07d207866d6ee2972eed439439a6
and give your opinion.

sashko:
@tolkien whatever works. Thanks.

gunnarx:
Thanks for this, Oleksandr, and on behalf of several people in GENIVI I apologize for slow processing.

I cherry-picked 4a64b49 from 12.0 to master (new id: b9d4aadb).  This is in the process of porting patches to better align all branches so that they all include relevant fixes).

Hopefully this works consistently.  If it does not, I will commit the change that disables man-generation instead.  I commented a lot more on it further up, so let's not repeat too much - but it boils down to this:  A bigger job should probably be done here to either include **all** relevant man-pages on the image, or perhaps **none** instead (and providing them elsewhere).

gunnarx:
Closing based on the above.

PR #57, opened 2017-08-09
node-state-manager 

frznlogic:
Don't use SRCREV as it collides with revs specified in SRC_URI in append


frznlogic:
The architecture/design in the node-state-manager is not very nice and we need to use a git repository inside the git repository. This is done in a .bbappend in the g-d-p with PR118, but SRCREV is clashing with using revs= in the git:// url. 

Is this an ok solution?

DunderRoffe (approved):

frznlogic:
Not needed. Closing PR.

PR #58, opened 2017-10-03
Include gstreamer and cairo appends in 13.0 branch

tobsan:
Trying to do a yocto build synced with pyro branches (for which I guess 13.0 is the closest match) does not work without these changes.

These changes were already in master, this is just a cherry-pick down to 13.0


tobsan:
Just noticed all of these are in #55 (so please merge that one instead). Sorry about that.

PR #59, opened 2017-11-20
Update README with new branch & support policies

gunnarx:
I asked @slawr for feedback on this before sending the PR.  

This should not be too surprising and mostly in line with expectations I think.  I'd like to get it fixed and communicated though.  Asking @GordanM, @sashko,  @frznlogic and others who have recently been sending PRs to also give feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,46 @@ Poky GENIVI compliant.
 Branch Policy
 -------------
 
-Please note:
-- The most recent development is done on the highest numbered branch
-- Master branch is used as a stable version, typically equal to the most recently *released* numbered branch
-- Maintenance patches are added to both master and released branches
+- New development is done on **master** branch and most new change requests
+(PRs) should be proposed as changes to master, unless you know they are
+applicable to a particular release only.
+- Somewhere near a new major release, a numbered release branch (e.g. 14.0)
+is created from master, and the new branch goes into stabilization/release phase.
+- Major numbers are stepped up every new release.  The release schedule is
+driven by a time plan.  Use [semantic versioning](http://semver.org/) for
+minor and patch numbers.
+- GENIVI-specific release tags, such as "M-0.2" for the second baseline
+*pre*release for the "M" platform may also be added.
+- After a versioned release, the release branch remains for maintenance and
+updates.
+- The project maintainer shall ensure that relevant patches are
+cherry-picked to every branch where they apply. I.e. patches should be
+back-ported *at minimum* to the Support Window versions as defined below.
+- In general, prefer a linear commit history (rebase and cherry-pick),
+  applying merge commits only where absolutely necessary to sort out a
+  complex merge situation (which we should rarely have).
+
+Support Window (Bugfixes, improvements)
+---------------------------------------
+
+- Because of available resources, and often low engagement from those companies
+that have settled on a version and gone into a production project, support
+is given for only the most recently released version, and *one* version
+before it.
+
+Support Window (Security)
+-------------------------
+
+- If critical security fixes are identified, the maintainer should apply
+them (if applicable) to the most recently released version and *two*
+versions before it.
+
+**Note** however that there is currently no quantified or documented
+commitment to tracking CVEs, nor any guarantee to apply all possible
+security fixes.  While it is of course tracked to the best of the
+maintainer's ability, the project is dependent on community input.  All the
+_responsibility remains on the adopting companies_ to secure their final
+products.
 
 Contribution
 -------------


README.md
```diff
@@ -8,10 +8,43 @@ Poky GENIVI compliant.
...
 
-Please note:
-- The most recent development is done on the highest numbered branch
-- Master branch is used as a stable version, typically equal to the most recently *released* numbered branch
-- Maintenance patches are added to both master and released branches
+- New development is done on **master** branch and most new change requests
+(PRs) should be proposed as changes to master, unless you know they are
+applicable to a particular release only.
+- Somewhere near a new major release, a numbered release branch (e.g. 14.0)
+is created from master, and the new branch goes into stabilization/release phase.
```
gunnarx: that new branch

README.md
```diff
@@ -8,10 +8,43 @@ Poky GENIVI compliant.
...
+- Release numbers use [semantic versioning](http://semver.org/), and are
+potentially also tagged with GENIVI-specific release tags, such as "M-0.2"
+for the second baseline prerelease for the "M" platform release.
+- After a versioned release, the release branch remains for maintenance and
```
slawr: Are you trying to describe the current numbering scheme or explain a new one? Doesn't the current numbering scheme not strictly follow semantic versioning, as the major number changes for each cycle regardless of whether it is back incompatible API or not?

As an aside I was never particularly fond of having both the A-x.y and x.y.z numbering scheme for the baseline. I thought it could be confusing for those not familiar with the release process, although I realise the A-x.y releases are there to facilitate the two pre-releases.

It can also be a little confusing when you look at the meta-ivi tags and find for example 12.0.0 and N-1.0 are not exactly equivalent due the former having a different distro version number. This might be an opportunity to tighten up the release process, i.e. the distro number should be bumped on N-1.0 so they are equivalent. That would be the subject of a separate task though of course. I may be forgetting some subtlety as to why that currently is though. 

Back on the PR: maybe we should just be more specific here and say there will be two A-x.y pre-releases, followed by x.y.z releases at the point of A-1.0. Alternatively it could point at something more specific on-line, but it feels like we ought to be able to do it here.

README.md
```diff
@@ -8,10 +8,43 @@ Poky GENIVI compliant.
...
+- Because of available resources, and often low engagement from those companies
+that have settled on a version and gone into a production project, support
+is given for only the most recently released version, and *one* version
+before it.
+
+Support Window (Security)
+-------------------------
+
+- Critical security issues should be fixed in *two* versions behind the most
+recently released.
```
slawr: Hopefully that will not be read as that critical security issues are being fixed two versions behind _for_ people. That some how it is falling on the maintainer/BIT shoulders to find/monitor security fixes and apply them.

I think that is what you try to address in the following paragraph. Making it more of a statement of how far back the project would normally be taking in fixes from the community based on current resources.

That said, I'm not sure how I might rewrite this section apart from possible inverting the text in some way. Maybe the problem is the "should" in the first para and what is could be interpreted as.

README.md
```diff
@@ -8,10 +8,43 @@ Poky GENIVI compliant.
...
+for the second baseline prerelease for the "M" platform release.
+- After a versioned release, the release branch remains for maintenance and
+updates.
+- The project maintainer shall ensure that relevant patches are
+cherry-picked to every branch where they apply. I.e. patches should be
+back-ported *at minimum* to the Support Window versions as defined below.
+- In general, prefer a linear commit history (rebase and cherry-pick),
+  applying merge commits only where absolutely necessary to sort out a
+  complex merge situation (which we should rarely have).
+
```
slawr, line 29: +1 on making expectations clear here

slawr (commented):
Generally some great solid improvements to describing the policy.
Some feedback below. Unfortunately I am not super clear on requested changes.

README.md
slawr: The semantic versioning clarification in the update deals with the concern I had

README.md
```diff
@@ -36,15 +38,15 @@ before it.
...
 **Note** however that there is currently no quantified or documented
 commitment to tracking CVEs, nor any guarantee to apply all possible
-security fixes. While it is of course tracked to the best of the
-maintainer's ability, the project is dependent on community input, and all
-the _responsibility remains on the adopting companies_ to secure their
-final products.
+security fixes.  While it is of course tracked to the best of the
+maintainer's ability, the project is dependent on community input.  All the
+_responsibility remains on the adopting companies_ to secure their final
+products.
```
slawr, line 51: Good clarifications in the update Gunnar.

slawr (approved):
With the clarifications in the update I approve the changes

README.md
gunnarx: Confirmed on today's BIT team meeting that we are in agreement and ready to merge this.

gunnarx:
Squashed, rebased & merged.

PR #61, opened 2018-01-18
Bump rocko

frznlogic:
This PR bumps meta-ivi to rocko release.


slawr:
Hi @frznlogic 

Good to see the rocko work is progressing.

The PR description doesn't say whether this is just showing visibility of a work in progress, or you think these commits are ready for review/merge. I know you said in the BIT call that your own fork was an early snapshot and needed some commit history rewriting. Maybe that applies here as well in the current PR state?

My initial comment is that commit messages are very basic, with most having a "headline" only and no body description. For example the vsomeip patch. There is nothing to give context to why it is needed or what it fixes. Such as a reference to a bug tracker ticket/issue or mailing list discussion. The patch itself has no signed off or reviewed by entries. That occurs sometimes of course when something comes from upstream, but then supporting information is even more required. This might be added as a temporary patch until it is taken upstream for example. Coming to this commit months or years down the line it could be hard to piece together why it exists.

frznlogic:
Hi,

I think the history is better in this branch than on the gdp branch to be honest, but yes, you are right, it's lacking details for some of the stuff. I'll be adding it. 

Long story: As for the vsomeip, it's a weird case, the patch was written by me as it fails to compile with the new compilers and I tried finding the maintainers but was unable to find them at that point, so I added the patch temporarily in meta-ivi to make it compilable again and then forgot about it.... 

Short question: Do you have a name/mail I can contact regarding vsomeip?

slawr:
>Long story: ..

Thanks for the background. As its your patch I would say add a signed off to the patch and expand the commit msg :)

>Short question: Do you have a name/mail I can contact regarding vsomeip?

With vsomeip I am not sure. I recall there being the two versions from Pelagicore and BMW so it may be a mixed team.  Reference list for maintainers is supposed to be the [project table](https://at.projects.genivi.org/wiki/display/PROJ/Projects+Home) in the Project wiki. That currently says:
>Juergen Gehring – Mail to genivi-maint-vsomeip@genivi.org

Try that. I believe that the OSS team were going through the list of maintainers. You could confirm with @jeremiah and @gunnarx.

frznlogic:
Thank you, I managed to find Juergen Gehring. I've sent a PR for vsomeip and added that patch here instead for the time being with commit msg etc :). 

I think pelagicore was at some point part of developing vsomeip but I'm not sure, there's at least no fork in pelagicore.. Thank you for the information however, I'll keep the project table in mind in the future :).

meta-ivi/recipes-extended/persistence-common-object/persistence-common-object_1.0.3.bb
```diff
@@ -20,4 +20,8 @@ inherit autotools-brokensep pkgconfig
 
 EXTRA_OECONF = " --with-database=key-value-store "
 
-FILES_${PN} += "${datadir}"
+do_install_append() {
+    rm -rf ${D}${sysconfdir}
```
gunnarx, line 24: Is this safe or could it interact with other packages?  
And why is it needed to do this instead of not touching it?

meta-ivi-test/recipes-extended/persistence-client-library/persistence-client-library/0001-memset-improperly-initialized-length-of-elements.patch
```diff
@@ -0,0 +1,309 @@
+diff --git a/test/persistence_client_library_test_file.c b/test/persistence_client_library_test_file.c
```
gunnarx: Me no like... ;-) Why don't we upstream the patch instead?  
(By the way, I have commit access to these repos, just in case the maintainer is not responsive, if that's the issue - and this change seems clear enough.)

meta-ivi/conf/distro/poky-ivi-systemd.conf
```diff
@@ -20,13 +20,13 @@ DISTRO_FEATURES_remove = "x11"
 DISTRO_FEATURES_BACKFILL_CONSIDERED = "sysvinit"
 
 # do not use gstreamer 1.2.3 by default
-PREFERRED_VERSION_gstreamer1.0              ?= "1.10.4"
```
gunnarx: The commit comment talks about 1.8.3 but this is 1.10.4?  (no big deal though - we could let that slip through)

meta-ivi/recipes-graphics/wayland/wayland-ivi-extension_2.0.bb
```diff
@@ -0,0 +1,25 @@
+SUMMARY = "Wayland IVI Extension"
```
gunnarx: Should you git rm also the old 1.13 recipe or do we keep it intentionally?

gunnarx (changes requested):

gunnarx:
Please push the vsomeip maintainer to include patch upstream if possible.  I didn't request it in review since I would (reluctantly) accept that patch, more than the  persistence change which I definitely think we should put in the component instead.  But it's a good test if vsomeip can accept quick changes.  If that is updating to new compilers, it should be a safe and good change to make.

commit comments, I think most are quite OK.  But pco: "another package" - which package?

frznlogic:
Patch for vsomeip seems to have been upstreamed. I've fixed the recipe to just bump versions instead.

meta-ivi/recipes-extended/persistence-common-object/persistence-common-object_1.0.3.bb
frznlogic: It should not interact with other packages as it's done on the installation destination dir for that specific package, ie, build/tmp/work/<architecture>/<package>/<version>/$sysconfdir

bitbake fails with an error if there are installed files that are not added to any FILES_x directives.

frznlogic: If you have a better solution I'm all ears :)

frznlogic: I guess the optimal solution would be if only one of the packages actually installed that file though :)

gunnarx (approved):

slawr:
Ah good you picked up on the meta-yocto renaming to meta-poky. I was going to raise it as a review comment as it had me wondering if you had tested your builds with [TEMPLATECONF](https://github.com/frznlogic/meta-ivi/blob/3e1319763c3273a5b8cedf98b2acc3eb79584928/README.build#L8) set.

gunnarx:
> Ah good you picked up on the meta-yocto renaming to meta-poky. I was going to raise it as a review comment as it had me wondering if you had tested your builds with TEMPLATECONF set.

That was the error we saw in CI so I had to check what was up.  Bitbake isn't great here - it just prints "ERROR: Unable to start bitbake server" and you have to dig into logs. :-(

slawr:
> That was the error we saw in CI so I had to check what was up. Bitbake isn't great here - it just prints "ERROR: Unable to start bitbake server" and you have to dig into logs. :-(

Weird I thought it coughed at the bitbake parse saying  the layer was unavailable.

Update: I just managed to actually try to kick off a build with the old name in bblayers.conf. I got the following good guidance directly in the terminal so I guess it is a foible of how the CI takes output:
$ time bitbake pulsar-image
ERROR: Unable to start bitbake server
ERROR: Server log for this session (/data/steve/genivi-14/m3sk/y-baseline/build/bitbake-cookerdaemon.log):
--- Starting bitbake server pid 11651 at 2018-02-13 17:34:25.162666 ---
ERROR: Layer directory '/data/steve/genivi-14/m3sk/y-baseline/poky/meta-yocto' does not exist! Please check BBLAYERS in /data/steve/genivi-14/m3sk/y-baseline/build/conf/bblayers.conf

frznlogic:
I'll pull this PR in, it builds and runs on all platform but the vexpressa9 which seems to be due to qemu/poky issues rather than us at the moment.

PR #62, opened 2018-01-23
node-startup-controller: Bump to 1.0.3

erikboto:
- Bump version
- Enable by default
- Add node-state-manager as a requirement

### DIFF
--- a/meta-ivi/recipes-extended/node-startup-controller/node-startup-controller_1.0.3.bb
+++ b/meta-ivi/recipes-extended/node-startup-controller/node-startup-controller_1.0.3.bb
@@ -10,29 +10,26 @@ HOMEPAGE = "https://www.genivi.org/"
 SECTION = "base"
 
 LICENSE = "MPLv2"
-LIC_FILES_CHKSUM = "file://COPYING;md5=815ca599c9df247a0c7f619bab123dad"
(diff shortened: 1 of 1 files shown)


meta-ivi/recipes-extended/node-startup-controller/node-startup-controller_1.0.3.bb
```diff
@@ -45,14 +42,3 @@ FILES_${PN} += "\
 FILES_${PN}-dbg += "\
     ${libdir}/${BPN}-1/.debug/*ler \
     "
-
-FILES_${PN}-nsm-dummy = "\
```
frznlogic, line 49: was the nsm dummy package removed upstream? Is there any other reason it was removed?

erikboto: It was only useful before NodeStateManager was done I guess, and maybe if you want to run component testing. But there's really no reason why it should be shipped on target.

It was also disabled from being built by default after NSM was done, see https://github.com/GENIVI/node-startup-controller/commit/fba1b26575c75390be28a8a020953f030114df87.

frznlogic: Hm, I see your point. So let's get rid of it then.... I'm accepting this PR.

frznlogic (approved):

gunnarx (approved):

PR #64, opened 2018-02-21
Add CI-build script

gunnarx:
Mostly untested, but it starts a build at least.

Proposal for discussion.  Likely needs a couple of updates before merging.

One concern is where the dependent layer are to be placed, but the current expectation seems to be that poky/ meta-openembeded/ and meta-gplv2/ should be adjacent to the meta-ivi "root" (repository working dir):

```
$ cd myworkdir
$ tree -L 2 -d
├── meta-gplv2
│   ├── ... (abbreviated)
├── meta-ivi
│   ├── build
│   ├── meta-ivi
│   ├── meta-ivi-bsp
│   ├── meta-ivi-test
│   └── scripts
├── meta-openembedded
│   ├── ... (abbreviated)
└── poky
│   ├── ... (abbreviated)
```

The script was tested as follows:

```
$ meta-ivi/scripts/ci-build.sh
```
or
```
$ cd meta-ivi
$ scripts/ci-build.sh
```
Both seem to work as they should.

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -234,7 +234,7 @@ trap cleanup SIGINT SIGTERM
 
 D=$(dirname "$0")
 cd "$D"
-BASEDIR=$(readlink -f "$PWD/..")
+BASEDIR=$(readlink -f "$PWD/../..") # One dir above meta-ivi!
 
 # The user must define this
 ensure_var_is_defined TARGET
@@ -324,12 +324,12 @@ echo "SSTATE_DIR = $DL_DIR"
 echo "STANDARD_RELEASE_BUILD" = "$STANDARD_RELEASE_BUILD"
 
 # INIT
-cd "$BASEDIR"
+cd "$BASEDIR/meta-ivi"
 echo "*** Initializing layers"
 scripts/checkout
 echo "*** Initializing conf"
 export TEMPLATECONF=$PWD/meta-ivi/conf
-source ../poky/oe-init-build-env build
+source ../poky/oe-init-build-env ../build
 
 # build steps
 # We are now in build/
@@ -353,7 +353,7 @@ if [[ "$KEEP_TMP" != "true" ]]; then
    rm -rf tmp
 fi
 
-cd "$BASEDIR"
+cd "$BASEDIR/meta-ivi"
 
 # Need to set an identity because if it is unset (which it could be in a CI
 # environment), some patching commands will fail the build.
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -1,38 +1,7 @@
 #!/bin/bash -e
 
-# (C) 2017 Gunnar Andersson
+# (C) 2018 GENIVI Alliance
 # LICENSE: MPLv2
-
-# User warning!
-
-# A casual user will need to read the script to understand what the first
-# parameter shall be to make it actually run:
-#
-
-# Fail script unless you give the magic value:
-if [[ "$1" != CI_FLAG ]]; then
-
-  cat <<EOT
-
-READ THIS: This script was not designed for personal use.
-
-If you still use it, you hereby agree to any applicable End-User License
-Agreement (EULA) of the corresponding BSPs that are used.  You must read
-and understand the obligations by studying the script.
-
-It is still a useful script but a few things could be destructive, e.g.
-"git reset --hard" if you use this in your development environment instead of
-a throw-away environment like CI.
-
-Please make sure you commit/backup local changes and read the script to
-understand what it does!
-
-After that I hope you enjoy using it.
-
-EOT
-
-   exit 2
-fi
 
 # ---- General settings ----
 
@@ -80,7 +49,7 @@ stop_immediately() {
 }
 
 append_local_conf() {
-  LOCAL_CONF="$BASEDIR/gdp-src-build/conf/local.conf"
+  LOCAL_CONF="$BASEDIR/build/conf/local.conf"
   if [[ -f "$LOCAL_CONF" ]]; then
     echo -n "Appending to local.conf: "
     cat <<EOT | tee -a "$LOCAL_CONF"
@@ -126,7 +95,7 @@ stage_artifact() {
 # LAYER_<layername>_TAG
 
 # *NOTE*  Variables cannot have dashes in name therefore, underscores must be
-# used!  For example: LAYER_meta_browser_COMMIT for meta-browser layer
+# used!  For example: LAYER_meta_openembedded for meta-openembedded layer
 
 # If an override variable is defined, build with this version of the layer
 # instead.  If multiple conflicting values, then the most specific override
@@ -200,23 +169,11 @@ print_layer_overrides() {
   done
 }
 
-# FIXME: Asking git to list submodules would be better
 LAYERS="
-meta-browser
-meta-erlang
-meta-genivi-dev
-meta-intel
-meta-iot-web
 meta-ivi
-meta-linaro
-meta-oic
-meta-openembedded
-meta-qcom
-meta-qt5
-meta-raspberrypi
-meta-renesas
-meta-rvi
 poky
+meta-gplv2
+meta-openembedded
 renesas-rcar-gen3
 "
 
@@ -252,6 +209,7 @@ stop_if_failure
 # The values can be overridden by defining environment variables
 # If no value given, use this default:
 define_with_default BUILD_SDK false
+define_with_default BUILD_TEST_IMAGE false
 define_with_default COPY_LICENSES false
 define_with_default LAYER_ARCHIVE false
 define_with_default CREATE_RELEASE_DIR false
@@ -276,15 +234,8 @@ define_with_default KEEP_TMP false
 
 stop_if_failure
 
-git_gdp="https://github.com/GENIVI/genivi-dev-platform"
+#git_gdp="https://github.com/GENIVI/meta-ivi"
 branch="master"
-
-# Special case: Use porter script to copy graphics drivers for koelsch
-if [[ "$TARGET" == "koelsch" ]]; then
-  GFX_MACHINE=porter
-else
-  GFX_MACHINE=$TARGET
-fi
 
 # cd workingdir
 MACHINE="$TARGET" # For most boards - exceptions handled below
@@ -322,6 +273,7 @@ echo "COMMIT = $COMMIT"
 echo "RELEASE = $RELEASE (currently unused)"
 print_layer_overrides "$LAYERS"
 echo "BUILD_SDK = $BUILD_SDK"
+echo "BUILD_TEST_IMAGE = $BUILD_TEST_IMAGE"
 echo "COPY_LICENSES" = "$COPY_LICENSES"
 echo "CREATE_RELEASE_DIR" = "$CREATE_RELEASE_DIR"
 echo "DL_DIR = $DL_DIR"
@@ -336,8 +288,16 @@ echo "SOURCE_ARCHIVE" = "$SOURCE_ARCHIVE"
 echo "SSTATE_DIR = $DL_DIR"
 echo "STANDARD_RELEASE_BUILD" = "$STANDARD_RELEASE_BUILD"
 
+# INIT
+cd "$BASEDIR"
+echo "*** Initializing layers"
+scripts/checkout
+echo "*** Initializing conf"
+export TEMPLATECONF=$PWD/meta-ivi/conf
+source ../poky/oe-init-build-env build
+
 # build steps
-cd "$BASEDIR/gdp-src-build"
+# We are now in build/
 
 # If DL/SSTATE are to be reused it is normally by the use of
 # REUSE_STANDARD_DL_DIR and REUSE_STANDARD_SSTATE_DIR.  Local dirs left in the
@@ -375,8 +335,6 @@ if [[ -n "$FORK" ]]; then
   git fetch
   # V *danger* V.  One reason why you should not use this script if it's not in CI
   git reset origin/master --hard
-  # After this, submodule sync/update will be done by init.sh
-  # this may need some further testing!
   git checkout $BRANCH  # <- note this should be ok even if $BRANCH is an empty value
 fi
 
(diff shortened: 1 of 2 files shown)


gunnarx:
I believe it works now.  Please review since there might be some old cruft left from GDP version it was based on.  
- If reading through it makes sense and fix what you can see
- If it builds the image correctly
... then it can be merged

slawr:
I need to leave early today but I'll try to kick off a local build for R-Car either today or tomorrow and will let you know the result.

scripts/ci-build.sh
```diff
@@ -0,0 +1,512 @@
...
+AGENT_STANDARD_SGX_LOCATION="/var/go/sgx_bin"
+AGENT_STANDARD_SGX_GEN3_LOCATION="/var/go/sgx_bin_gen3"
+
+# ---- Helper functions ----
+
+fail=false
+
+# This is some trickery, but quite useful.
+# Return the value of a variable whose name is stored in another variable
+deref() { eval echo \$$1 ; }
```
jeremiah: Why not just call ${other_variable}?

scripts/init.sh
```diff
@@ -0,0 +1,33 @@
...
+echo "Checking out the right versions of layers"
+echo poky:
+scripts/checkout_layer_hash.sh poky
+echo meta-openembedded:
+scripts/checkout_layer_hash.sh meta-openembedded
+
+source poky/oe-init-build-env
+
+echo
+echo "Now run:  bitbake miranda-image"
```
jeremiah: I see this is just a comment but it says "miranda" when other scripts say "pulsar". Does this matter? Will it confuse someone (like it confuses me. :)

scripts/ci-build.sh
gunnarx: ${var} returns the value stored in var
deref var returns the value stored in the variable whose name is stored in var
e.g.
```
x=y
y=z
echo $x    <- prints out "y"
echo $(deref x)    <- prints out "z"
```

gunnarx: also
```
x=y
y=z
echo ${$x}  <- you'd think this should work, but...
--> bash: ${$x}: bad substitution
```
But maybe there's another way...

scripts/init.sh
gunnarx: This (and all other files) should not be part of PR - only ci-build.sh was intended.
I must have goofed on the last commit.  I'll fix and update.

scripts/ci-build.sh
```diff
@@ -0,0 +1,512 @@
...
+  done
+}
+
+# Because of the order they are called we can't use output from the
+# layer_override function above, so we have some repeated code here.
+print_layer_overrides() {
+  local names="$1" name clean_name suffix var_name value d
+
+  for name in $names ; do
+    clean_name=$(echo "$name" | sed 's/-/_/g')
```
gunnarx: Could more succinctly use string substitution directly in bash, it's just that I'm familiar with the sed syntax...

scripts/ci-build.sh
```diff
@@ -0,0 +1,512 @@
...
+# If an override variable is defined, build with this version of the layer
+# instead.  If multiple conflicting values, then the most specific override
+# (commit) wins
+
+layer_override() {
+  local name=$1 suffix var_name value d
+
+  # Variables cannot have dashes in name therefore, underscores must be
+  # used.  For example: LAYER_meta_browser_COMMIT for meta-browser layer
+  clean_name=$(echo "$name" | sed 's/-/_/g')
```
gunnarx, line 141: Could use string substitution directly in bash, it's just that I'm familiar with the sed syntax...

gunnarx:
> This (and all other files) should not be part of PR - only ci-build.sh was intended. I must have goofed on the last commit. I'll fix and update.

The image definition files are now gone.

gunnarx:
Build script seems to work now, according to [this test pipeline](https://go.genivi.org/go/tab/build/detail/meta-ivi-with-script/13/Build/1/init_and_bitbake).  (A few comments were removed after that, but it should be the same script)

Might be some corner-case bugs remaining of course. 

Some instructions to copy artifacts to a staging/ directory should be added but I'll let @frznlogic consider how to add that later.  

I recommend merging now.

frznlogic:
Hi, just saw that you mentioned it should be pulled in now.

Some general comments/questions. 
Wouldn't it be better to break the scripts into their own repository for code reusability and to avoid duplication of efforts? Ie, a separate repository containing this script modified to be more general,  provide configurations within the repo or the go.cd environment for what to fetch etc. 

The bash code is a bit under-documented, some comments on the intention of the functions would be nice. Please explain the trickery and the rationale instead of calling it trickery also :)

frznlogic:
Also, please remove the WIP in the PR name :)

gunnarx:
> Wouldn't it be better to break the scripts into their own repository for code reusability and to avoid duplication of efforts? 

You mean the script used by meta-ivi and GDP?  Yes, there is a bit of duplication between them, but I felt the argument to keep them with the source (below) stronger so far.  I am also unsure how different GDP & meta-ivi scripts will become, or continue to be.  (Related to the strategy of parent-project for meta-ivi, with submodules, etc.) so this is a start, with potential for refactoring later.

> Ie, a separate repository containing this script modified to be more general, 

Basically my argument against this has been that ci-build.sh *can* be usable also for advanced users (and I hope we can take any steps to improve that usability, such as documentation).   For me that points towards keeping the build-script with the source that users already have, instead of requiring users to pick up another new repository.  We also used to have that, a "baseline-ci-helper" repo for Go builds, but it was just overhead...

There are a few dependencies between script & source also (for example the list of sub-layers inside the script).  When there's a dependency that needs to be in sync with eachother, it points towards keeping it in one repo if possible.  There's the possibility of a "scripts" submodule perhaps, but I don't know if we are there yet.

> provide configurations within the repo or the go.cd environment for what to fetch etc.

I find I often prefer a straight-forward approach, using git branches and such, to manage differences, instead of creating too much structure.  Unless the problem/project is big enough to actually gain speed in the long run by creating it.  For this you need to decide on some "configuration language" of sorts, then keep those configurations in sync with each project, and so on.
I'm not sure this build stuff big enough  to warrant the effort.  BUT, if you envision something larger here, like reuse in pelux as well then I might be convinced I guess. I'm open to suggestions.

> The bash code is a bit under-documented, some comments on the intention of the functions would be nice. 

Agreed, but could it be future improvement?  (I was hoping you take over the script, assuming it gets merged here...).   If you review-comment on the lines that you feel must be fixed, I'll address those.
**EDIT** Added comments in bf5e8d3

> Please explain the trickery and the rationale instead of calling it trickery also :)

In my opinion, the comment just above the deref function explains what the function does as clearly as can be done.  My answer to Jeremiah basically repeated the comment. The need for the function can be seen where it is used, but OK, I'll add rationale.

gunnarx:
@frznlogic Please review latest commit for more documentation.  We could squash before merge if you prefer.

scripts/ci-build.sh
gunnarx: The reason something like it is needed here is that the variable *name* itself is generated by the program (it's not known, like in the above example of x & y).

slawr:
>>     Ie, a separate repository containing this script modified to be more general,
> 
> Basically my argument against this has been that ci-build.sh can be usable also for advanced users (and I hope we can take any steps to improve that usability, such as documentation). For me that points towards keeping the build-script with the source that users already have, instead of requiring users to pick up another new repository.

If its simple to use I could see that having it in the source may help contribution, including component maintainers, if it simplifies setup of the baseline when they are not used to doing it.

gunnarx:
Well, yes the script is simple(r) to use than the alternatives.  Currently the script checks out the right commits (taken from README), but does *not* clone the layers themselves.  I think that should be added, unless we have some other parent-project setup.

So as of now: these are basically the minimal steps you must do:
```bash
$ git clone git://github.com/GENIVI/meta-ivi
$ git clone git://git.yoctoproject.org/poky
$ git clone git://git.openembedded.org/meta-openembedded
$ git clone git://git.yoctoproject.org/meta-gplv2
$ export TARGET=qemuarm64
$ meta-ivi/scripts/ci-build.sh
```
That should start the build.  It still takes some steps away.  Steps 2-4 ought to be eliminated in some way - please let me know how you think it should be done.

slawr:
Thanks for the note and sorry for not trying the script earlier like I said I would. It slipped my mind with all the Domain Interaction work recently. The note makes it clear where I would do the extras for a community board.

When I think of baseline builds I see two broad use cases, building the "as is" baseline and the baseline as part of something larger. The former could be for the official qemu or community targets. Community targets might mean the addition of a layer or two but you are not radically reshaping the distro. For me they are the principal targets of such "helpers". It may simply be that the helper needs to be flexible enough to support the community targets either by extension or manual local work to add things.

Personally I would not worry too much about the second use case, baseline as part of something much larger, currently. Such a distro is likely to have its own tooling anyway.

scripts/ci-build.sh
```diff
@@ -0,0 +1,562 @@
...
+# If no value given, use this default:
+define_with_default BUILD_SDK false
+define_with_default BUILD_TEST_IMAGE false
+define_with_default COPY_LICENSES false
+define_with_default LAYER_ARCHIVE false
+define_with_default CREATE_RELEASE_DIR false
+define_with_default MIRROR "https://docs.projects.genivi.org/releases/yocto_mirror"
+define_with_default PREMIRROR ""  # By default none (but we have the shared DL_DIR)
+define_with_default RM_WORK false
+define_with_default REUSE_STANDARD_DL_DIR true
```
frznlogic: I'd prefer the default for DL_DIR to be false

scripts/ci-build.sh
```diff
@@ -0,0 +1,562 @@
...
+define_with_default REUSE_STANDARD_SSTATE_DIR true
```
frznlogic: I'd prefer the default for SSTATE_DIR to be false. 

It's easy to set this for the CI build and generates a fairly complex error message for whatever user right now.

scripts/ci-build.sh
```diff
@@ -0,0 +1,562 @@
...
+# (C) 2018 GENIVI Alliance
+# LICENSE: MPLv2
+
+# ---- General settings ----
+
+# Fixed cache location on all agents that support this.
+AGENT_STANDARD_DL_DIR="/var/cache/yocto/downloads"
+AGENT_STANDARD_SSTATE_DIR="/var/cache/yocto/sstate"
+AGENT_STANDARD_SGX_LOCATION="/var/go/sgx_bin"
+AGENT_STANDARD_SGX_GEN3_LOCATION="/var/go/sgx_bin_gen3"
```
frznlogic, line 12: All of these paths are specific to the CI environment on go.genivi.org. I'd prefer them to be either be tied to $HOME or $PWD if possible and then change them locally in the CI environment.

scripts/ci-build.sh
```diff
@@ -0,0 +1,562 @@
...
+}
+
+# Append a given value to the project's bitbake local.conf file
+append_local_conf() {
+  LOCAL_CONF="$BASEDIR/build/conf/local.conf"
+  if [[ -f "$LOCAL_CONF" ]]; then
+    echo -n "Appending to local.conf: "
+    cat <<EOT | tee -a "$LOCAL_CONF"
+$1
+EOT
```
frznlogic: The script will continue appending to an existing local.conf file for every time you try to run it. Ie, we wind up with the MIRRORS_append once every time we try to run this script as it is now.

scripts/ci-build.sh
```diff
@@ -0,0 +1,562 @@
...
+  fi
+}
+
+# ---- Main program ----
+
+trap cleanup SIGINT SIGTERM
+
+D=$(dirname "$0")
+cd "$D"
+BASEDIR=$(readlink -f "$PWD/..")
```
frznlogic: For some reason we wind up building in <path>/meta-ivi/build/... I'd prefer to keep the build directory in <path>/build, the same level as the other meta-layers.

frznlogic (changes requested):
Looks better with the comments etc. I've tried running the script and had some more minor issues with the defaults basically. What it boils down to imho, it would be nicer to move the burden of handling defaults from the users to the CI environment.

scripts/ci-build.sh
gunnarx: ... patches welcome :)

scripts/ci-build.sh
gunnarx: OK, sure, that could be changed.  This is a remnant from GDP where build/ and scripts/ are in parallel (since GDP git is a parent project but meta-ivi is just a layer).

scripts/ci-build.sh
gunnarx: OK

scripts/ci-build.sh
gunnarx: OK

scripts/ci-build.sh
```diff
@@ -0,0 +1,562 @@
...
+rm -f logs.tar logs.tar.gz
+find build/tmp/work \( -name "*.log" -o -name "log.*" -o -name "run.*" \) -print0 | xargs -0 tar uf logs.tar || true
+gzip logs.tar || true
+
+# The following will be copied/moved to the staging directory, but only if
+# the files exist.  It will silently continue, with any files that do exist.
+set +e  # Allow failures temporarily
+rm -rf staging
+shopt -s nullglob
+stage_artifact mv build/tmp/deploy/licenses
```
gunnarx, line 508: By the way, there isn't very deep thought behind the selection of move vs copy here.  Mostly I guess the idea should be to move large things instead of copy, to save disk space and time but to copy things like logs that are expected to be found when poking around the build tree...

All of this is sort of optimized for CI and we might want to put a condition on it.

scripts/ci-build.sh
```diff
@@ -0,0 +1,562 @@
...
+define_with_default COPY_LICENSES false
+define_with_default LAYER_ARCHIVE false
+define_with_default CREATE_RELEASE_DIR false
+define_with_default MIRROR "https://docs.projects.genivi.org/releases/yocto_mirror"
+define_with_default PREMIRROR ""  # By default none (but we have the shared DL_DIR)
+define_with_default RM_WORK false
+define_with_default REUSE_STANDARD_DL_DIR true
+define_with_default REUSE_STANDARD_SSTATE_DIR true
+define_with_default SGX_DRIVERS $AGENT_STANDARD_SGX_LOCATION
+define_with_default SGX_GEN_3_DRIVERS $AGENT_STANDARD_SGX_GEN3_LOCATION
```
gunnarx: Note that once the default for "Use standard dirs" is false, the user will be able to specify their own values for SGX bin, DL_DIR an SSTATE_DIR.

define_with_default() will not change a variable if it already has a value in environment.

scripts/ci-build.sh
gunnarx: These have the prefix AGENT_ because they are for the build agents.  They affect the user build only depending on REUSE_*, and if the user does not set its own SGX_DRIVERS location, etc.
So I propose to keep them.  We could set a user-friendly default, but what would that be?
Proposal?  $HOME/something?

scripts/ci-build.sh
gunnarx: Agreed, this is because CI environments are thrown away frequently.  So this could be improved for end-user, yes.

gunnarx:
Squashed one commit that was superfluous.  Question now is if we squash more or keep this logical history.
@frznlogic I'm not sure if your review is still valid or can be approved now?  I think you need to try the script and decide if it's OK.

frznlogic (approved):

frznlogic:
@gunnarx I'm good with it, let's get this in imho. Waiting for the builds to finish and then pulling it in :)

PR #60, opened 2017-12-06
Add "hardening" compiler flags

gunnarx:
NOTE: Untested - pushing PR for build test.

These gcc flags are quite standard practice but not always applied by
each and every component default settings - hence each Linux distro does
it in their own packaging.

It was requested by Phong Tran in PR #98 for GDP.  These flags are
applied to the whole distro however, which is why they are placed in
meta-ivi distro definition instead.

[GDP-754] Compiler Hardening Flags

Signed-off-by: Gunnar Andersson <gandersson@genivi.org>

PR #67, opened 2018-03-27
Bump dependencies: poky, meta-oe, meta-gplv2

frznlogic:
All three layers have been bumped for testing and getting the latest updates.


frznlogic:
@gunnarx do you mind taking an extra look at the last commit, I cherry picked your commit and added some "blacklists" for the security flags until it compiled basically. Is it ok with you?

Clarification, I basically wanted to build it with my current code base to see that it worked with the P-1.0 updates instead of the P-0.1. It made more sense to have the commit here instead.

gunnarx:
@frznlogic Yeah that's fine.  We can close #60 then.  (Before we do, just take note of my last comment/question on #60 and check that at some point in the future).

PR #69, opened 2018-05-04
qemu_2: Remove dependency on libusb

frznlogic:
This is a really old dependency append it's causing issues in builds
relying on the meta-ivi layer. Removing it as it doesn't seem necessary.


frznlogic:
@thenor can you test this change?

slawr:
So USB still works in qemu without it?

PR #70, opened 2018-05-18
distro: poky-ivi-systemd: inherit from poky

mirzak:
Build tested with

    export MACHINE=qemux86-64 ; bitbake pulsar-image

and boot tested with

    runqemu nographics

No obvious problems. This might expose some QA warnings and errors due to it being a bit different to what was present and what is in "rocko" poky.conf.

### DIFF
--- a/README.build
+++ b/README.build
@@ -52,16 +52,3 @@ to use am 7.0, put following lines to <build directory>/conf/local.conf
 # use audiomanager 7.0 until am7.4 issues are fixed.
 PREFERRED_VERSION_audiomanager          ?= "7.0"
 PREFERRED_VERSION_audiomanagerplugins   ?= "7.0"
-
-
-Build s/w packages using gstreamer1.0-\*\_1.2.3.bb instead of default version
------------------------------------------------------------------------------
-you may encounter a build error like below:
-
-ERROR: Nothing PROVIDES 'tremor' (but /opt/genivi-baseline_13/poky/../meta-ivi/meta-ivi/recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_1.2.3.bb DEPENDS on or otherwise requires it)
-ERROR: Required build target 'gstreamer1.0-plugins-base' has no buildable providers.
-Missing or unbuildable dependency chain was: ['gstreamer1.0-plugins-base', 'tremor']
-
-then you may need "tremor" package for build this package.
-add following line to your conf/bblayser.conf.
-BBLAYERS_append := " <your topdir>/meta-openembedded/meta-multimedia"
--- a/meta-ivi/conf/distro/include/default-providers-ivi.inc
+++ b/meta-ivi/conf/distro/include/default-providers-ivi.inc
@@ -3,4 +3,3 @@
 #
 PREFERRED_PROVIDER_udev ?= "systemd"
 PREFERRED_PROVIDER_udev-utils ?= "systemd"
-PREFERRED_PROVIDER_node-state-manager ?= "node-state-manager"
--- a/meta-ivi/conf/distro/include/default-providers-ivi.inc
+++ b/meta-ivi/conf/distro/include/default-providers-ivi.inc
@@ -1,5 +1,0 @@
-#
-# Default virtual providers
-#
-PREFERRED_PROVIDER_udev ?= "systemd"
-PREFERRED_PROVIDER_udev-utils ?= "systemd"
--- a/meta-ivi/conf/distro/include/default-providers-ivi.inc
+++ b/meta-ivi/conf/distro/include/default-providers-ivi.inc
@@ -4,9 +4,3 @@
 PREFERRED_PROVIDER_udev ?= "systemd"
 PREFERRED_PROVIDER_udev-utils ?= "systemd"
 PREFERRED_PROVIDER_node-state-manager ?= "node-state-manager"
-
-#
-# Default jpeg provider
-#
-PREFERRED_PROVIDER_jpeg ?= "jpeg"
-PREFERRED_PROVIDER_jpeg-native ?= "jpeg-native"
--- a/meta-ivi/conf/distro/include/default-providers-ivi.inc
+++ b/meta-ivi/conf/distro/include/default-providers-ivi.inc
@@ -1,7 +1,6 @@
 #
 # Default virtual providers
 #
-PREFERRED_PROVIDER_virtual/arm-oe-linux-gnueabi-depmod ?= "kmod-cross"
 PREFERRED_PROVIDER_udev ?= "systemd"
 PREFERRED_PROVIDER_udev-utils ?= "systemd"
 PREFERRED_PROVIDER_node-state-manager ?= "node-state-manager"
--- a/meta-ivi/conf/distro/poky-ivi-systemd.conf
+++ b/meta-ivi/conf/distro/poky-ivi-systemd.conf
@@ -3,7 +3,7 @@ require conf/distro/poky.conf
 DISTRO = "poky-ivi-systemd"
 DISTRO_NAME = "Yocto GENIVI Baseline (Poky/meta-ivi)"
 DISTRO_VERSION = "14.0.0"
-DISTRO_CODENAME ="pulsar"
+DISTRO_CODENAME = "pulsar"
 
 MAINTAINER = "meta-ivi <genivi-meta-ivi@lists.genivi.org>"
 
--- a/meta-ivi/conf/distro/poky-ivi-systemd.conf
+++ b/meta-ivi/conf/distro/poky-ivi-systemd.conf
@@ -42,14 +42,6 @@ SECURITY_LDFLAGS_pn-v86d = ""
 SECURITY_CFLAGS_pn-gettext = ""
 SECURITY_LDFLAGS_pn-gettext = ""
 
-# do not use gstreamer 1.2.3 by default
-PREFERRED_VERSION_gstreamer1.0              ?= "1.12.2"
-PREFERRED_VERSION_gstreamer1.0-plugins-bad  ?= "1.12.2"
-PREFERRED_VERSION_gstreamer1.0-plugins-base ?= "1.12.2"
-PREFERRED_VERSION_gstreamer1.0-plugins-good ?= "1.12.2"
-PREFERRED_VERSION_gstreamer1.0-plugins-ugly ?= "1.12.2"
-PREFERRED_VERSION_gstreamer1.0-libav        ?= "1.12.2"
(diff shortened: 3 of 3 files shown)


frznlogic:
Hi @mirzak, 

Looks nice, much cleaner code etc. I'll do some tests on it and see if everything works as intended etc.

slawr (commented):
Hi,

Looks like a good idea.

What was the thinking for eliminating the default providers include?

Just an aside but if you are reforming the distro setup I wonder if it would be a good idea to move the GPLv3 exclusion that is currently in the local.conf.sample [1] into the setup. 

[1] https://github.com/GENIVI/meta-ivi/blob/2c11b36653a3c5c0657dedfdfcf78278a9920f22/meta-ivi/conf/local.conf.sample#L82

Regards

Steve

mirzak:
> What was the thinking for eliminating the default providers include?

My reasoning behind removing them was that the file has not changed in 4 years, and what I can see there are only single providers for the packages that where listed there. The systemd udev parts where moved to the poky-ivi-systemd file instead.

I do not know the full background so I might be doing something wrong :). But unless you feel that there might be multiple providers added for those packages in a near by future I do not see a reason to keep them around.

In the end it is less "lines" to maintain, and has no real effect on the output at the moment.

> Just an aside but if you are reforming the distro setup I wonder if it would be a good idea to move the GPLv3 exclusion that is currently in the local.conf.sample [1] into the setup.

Makes sense to me.

slawr:
>>     What was the thinking for eliminating the default providers include?
> 
> My reasoning behind removing them was that the file has not changed in 4 years, and what I can see there are only single providers for the packages that where listed there. The systemd udev parts where moved to the poky-ivi-systemd file instead.
> 
> I do not know the full background so I might be doing something wrong :). But unless you feel that there might be multiple providers added for those packages in a near by future I do not see a reason to keep them around.
> 
> In the end it is less "lines" to maintain, and has no real effect on the output at the moment.

I don't have any historical input for the entries. For why the jpeg provider was added for example. I was thinking more of the file itself. The poky meta has a structure of splitting certain variables into [logical includes ](https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/conf/distro/include?h=rocko) such as the default providers, which meta-ivi appeared to be following. Which made my wonder about your reasoning for moving away from that. It's not a major issue, hence why I just added a comment rather than a change request. I don't think there is a right or wrong. When there are only a few variables it can be argued to have them all in a single file (the conf) as you moved towards or split into logical includes.

mirzak:
Mostly because there was so little to separate. Logical includes makes sense, but not really when they only include one or two lines. As you already stated as well.

But we could move it around a bit more, that is move the gstreamer preferred to the include file. Though I would like to question why the gstreamer preferred is needed as there is no multiple providers here either. And generally gstreamer stuff is "bsp" anyway. While we are at it :)

slawr:
> Mostly because there was so little to separate. Logical includes makes sense, but not really when they only include one or two lines. As you already stated as well.

I don't feel strongly it should be changed back. There are more important changes to be done.

> But we could move it around a bit more, that is move the gstreamer preferred to the include file. Though I would like to question why the gstreamer preferred is needed as there is no multiple providers here either. And generally gstreamer stuff is "bsp" anyway. While we are at it :)

My understanding is that gst preferred version vars came in when meta-ivi provided support for Gst 1.2.x for an earlier generation of SoCs. Poky meta moved on in version, but many of the SoCs still used Gst 1.2.x. To avoid a lot of duplication meta-ivi carried the Gst 1.2.x recipes but that required the default version to be set to what poky meta used. The Gst 1.2.x recipes were dropped when later generations of SoCs were available. The preferred version would have been a candidate for removal at the same time and to my mind still is, but it may have been kept for some reason...

gunnarx:
+1 on the patch

tolkien:
the reason why there is gst preferred version is to keep compatibility of legacy bsp while I maintained.
if there is no bsp in meta-ivi needs "old" gst, please, remove it and
notify that there is no longer support at release note.

mirzak:
It does not even point to an "old" gstreamer (based on @tolkien comment) anymore and just referees to the latest in poky. Which should make it safe to remove IMO.

Updated the pull-request, which now also drops the gstreamer stuff

slawr:
>the reason why there is gst preferred version is to keep compatibility of legacy bsp while I maintained.
if there is no bsp in meta-ivi needs "old" gst, please, remove it and
notify that there is no longer support at release note.

Thanks for the insight Tolkien. I hope things are well with you.

>It does not even point to an "old" gstreamer (based on @tolkien comment) anymore and just referees to the latest in poky. 

The logic was the other way around. It always pointed at the poky version by default. Then if a BSP wanted to use the gst 1.2.x recipes that were being carried then the default could be overwritten.

That doesn't change the outcome though. It can be removed :)

mirzak:
@frznlogic, any progress on the testing?

gunnarx:
I mean Oscar...
@tolkien - thanks for checking this.  Please also answer Oscar's email about git.yoctoproject.org :)
Any question?  Email me or Oscar, please.

mirzak:
Friendly ping

slawr (commented):
The commit msg states:
>While at it I did some cleaning, mostly variables that did not have actual effect on build output.

Given the important of the distro conf does this need more explanation? Down the line people may be wondering why the changes were made.

Should the clean up be split into a separate commit from the refactoring?

Regarding the gst changes I noticed today that there is a comment about selecting gst 1.2.x in one of the readme files. If gst 1.2.x is now fully removed then it looks like that comment could also be removed as part of the clean up. Link:
https://github.com/GENIVI/meta-ivi/blob/8efbab9130de3fedd62ef5ac6ad6fe138df610d9/README.build#L57

mirzak:
Pushed a new version where I have split up the pull-request and also updated the README.build based on your comment.

mirzak:
Any more comments? @frznlogic @slawr

slawr:
>Any more comments? @frznlogic @slawr

I think the reasoning behind the removals in the commit msgs is much more clearly explained.

Looking through them my only comment is on a514c764. The msg argues that machine overrides should reside in YBSP layers not the distribution configuration file. It goes on to remove them along with some other similarly h/w related values, but does not appear to redistribute them into meta-ivi-bsp. Of course you can't do that for all possible machines but meta-ivi has four officially supported qemu machines, x86, x86-64, arm64 and vexpressa9 (arm32) so it could be done for those.

Of course related to that is the important question of whether those settings are just historical and no longer needed? On that I don't have an opinion without further investigation.

Just a comment, rather than a change request.

mirzak:
I think it is safe to remove, https://github.com/GENIVI/meta-ivi/commit/2f70439bd9e15d744766bad34d2ae555ae441ce9

It was added in 2012, with the reverse of my logic of removing it :)

Edit: updated commit message though to be more clear

slawr:
Thanks for the historical commit. If the removal (the question of removal vs a move to a different layer) doesn't result in an unwanted loss of functionality then I don't see any objections.

btw thinking about something related to this came across this [package group ](https://github.com/GENIVI/meta-ivi/blob/master/meta-ivi/recipes-yocto-ivi/packagegroups/packagegroup-xserver-ivi.bb) for xserver in meta-ivi which also sets the XSERVER variable.

gunnarx:
There have been some fixes and updates since I reviewed.  Again I add my +1.
What is holding up the merge?

gunnarx:
(Let me know if you prefer more build pipelines to confirm the different hardware targets)

slawr:
> What is holding up the merge?

No objection from me for merge. Latterly I only had a comment that I hoped that the gfx removals did not break gfx for the meta-ivi official qemu targets. By break gfx I mean a simple sanity test like start weston and run the ivi-extension mock navigation.

mirzak:
My understand is that the reference images that you provide in meta-ivi are headless, that is no graphics? How would you test this?

Also XSERVER is Xorg specific configurations which is a completely different graphical back-end then wayland/weston. If XSERVER has any effect on wayland/weston that would be a _major_ bug in Yocto/OE-core and not in our layers.

I would argue that you do not even use XSERVER today as my understanding is that you build with wayland/weston and not Xorg (X11)

slawr:
Sorry @mirzak replying slipped off my task list.

> My understand is that the reference images that you provide in meta-ivi are headless, that is no graphics?

Yes meta-ivi provides no HMI. Part of the reasoning being that HMI is a differentiator for adopters, and meta-ivi is designed to allow it to be a base layer for other platforms so having to remove a reference HMI to use it in that way would be problematic.

That said, Genivi has been involved in the creation of gfx framework components because of the specialist requirements of IVI. First with the Layer Management component, which then was superseded by the Compositor component of which Wayland IVI-Extension is a part. One goal of meta-ivi is to prove that the Genivi Compliance can be built and integrated. So it includes recipes for the Compositor component reference implementation which includes Wayland/Weston and IVI-Extension. It's also integrated into meta-ivi-test. 

> How would you test this?
I'm not sure if you are asking in the general sense of how do you test gfx in meta-ivi without a HMI or more specifically about testing any functional loss with this PR.

Personally, I focus on Renesas h/w. A simple sanity test is to build the baseline with weston-init and the R-Car gfx packages added to the image. I can then boot it, start weston via systemd, use the ivi-extension support utility layer-add-surfaces to create a layer/surface and then run EGLMockNavigation, an ivi-extension enabled sample app that is included in the meta-ivi build, to display some gfx on it.

My belief is that previous baseline maintainers did similar for the official qemu targets. The sanity test I mentioned is detailed in [Test 1](https://at.projects.genivi.org/wiki/display/PROJ/Test+of+wayland-ivi-extension) in the ivi-extension section of the meta-ivi-test wiki pages before Oscar moved some of the info into the meta-ivi-test source. I can't comment on the qemu specifics as I have not run the qemu gfx in a long time. The deprecated test description also manually starts weston when I would use systemd to do it, but the bones of it are there.

So that is the background to my observations about qemu gfx still working. Although there is an assumption that it did in the recent past.

> 
> Also XSERVER is Xorg specific configurations which is a completely different graphical back-end then wayland/weston. 

>If XSERVER has any effect on wayland/weston that would be a major bug in Yocto/OE-core and not in our layers.
> 
> I would argue that you do not even use XSERVER today as my understanding is that you build with wayland/weston and not Xorg (X11)

Agreed. I expect it may be left from the days of the earlier Layer Management component but I am not sure. 

BTW we discussed the open meta-ivi PRs in the BIT call during Oscar's holiday. The simple readme changes are merged and this will be as well once Gunnar gets to it.

gunnarx:
Hardening flags and/or compiler updates causing a few issues with some components in GDP, but I would say the problem is there, not here.  Rebased&Merged.

PR #73, opened 2018-06-13
ci-build: Fix the git diff reporting on conf files

gunnarx:
See commit comment for details.


gunnarx:
@frznlogic Now the patch is OK I think.  
If you're OK to merge it, then the pipeline for CI checking dependent layers will start working as it should.

frznlogic:
LGTM.

PR #76, opened 2018-07-25
README.md: bump dependencies for sumo

sashko:


gunnarx:
To make the build test work, the changes need to be on the same PR. 
EDIT: I updated #75 instead and added the README to it.  So let's close this one..

PR #77, opened 2018-07-27
checkout_layer_hash.sh: Fetch selected branch

gunnarx:
The CI system will typically make shallow clones, and it might be set to
fetch a specific layer branch.  In a transition period, when the README
specifies a new branch we might lack the requested commit hash, but
since the README lists the branch name, we can use it and make sure it
is fetched first.

### DIFF
--- a/scripts/checkout_layer_hash.sh
+++ b/scripts/checkout_layer_hash.sh
@@ -77,7 +77,7 @@ EOT
 
    # Get value for item "what" (what = revision, branch, ...)
    value=$(
-   cat <<EOT | grep "$what" | awk '{print $2}'
+   cat <<EOT | fgrep "$what" | awk '{print $2}'
    $layer_info
 EOT
    )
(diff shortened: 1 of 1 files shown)


scripts/checkout_layer_hash.sh
```diff
@@ -77,7 +77,7 @@ EOT
 
    # Get value for item "what" (what = revision, branch, ...)
    value=$(
-   cat <<EOT | grep "$what" | awk '{print $2}'
+   cat <<EOT | fgrep "$what" | awk '{print $2}'
```
gunnarx, line 80: Just a minor fix to possibly avoid some unexpected regexp

gunnarx:
I think this could be merged @frznlogic ?

scripts/checkout_layer_hash.sh
frznlogic: Would you like this fix to be merged? Most of the PR was already merged with the 14.x-sumo PR.

frznlogic:
Merged!

PR #78, opened 2018-08-08
bluez5: delete obsolete patch

sashko:
Deleted patch since the change has been mainstreamed.


frznlogic:
This patch is not obsolete, it's fuzzed but still applies. I've fixed it in the latest 14.x-sumo commits today.

frznlogic:
@sashko please verify if the new patch applies cleanly etc?

PR #81, opened 2018-08-27
dbus.bbappend: Remove patch messing with readdir

gunnarx:
Let's test if this builds...

### DIFF
--- a/meta-ivi/recipes-core-ivi/dbus/dbus/build_fix-ubunt1604-qemux86.patch
+++ b/meta-ivi/recipes-core-ivi/dbus/dbus/build_fix-ubunt1604-qemux86.patch
@@ -1,24 +1,3 @@
---- dbus-1.10.10/dbus/dbus-sysdeps-unix.c.orig	2017-01-26 20:39:27.868191644 +0900
-+++ dbus-1.10.10/dbus/dbus-sysdeps-unix.c	2017-01-26 20:43:15.708197478 +0900
-@@ -4320,13 +4320,14 @@
-     {
-       for (;;)
-         {
--          struct dirent buf, *de;
--          int k, fd;
-+          struct dirent *de;
-+          int fd;
-           long l;
-           char *e = NULL;
- 
--          k = readdir_r (d, &buf, &de);
--          if (k != 0 || !de)
-+          errno = 0;
-+          de = readdir (d);
-+          if (!de && errno)
-             break;
- 
-           if (de->d_name[0] == '.')
 --- dbus-1.10.10/dbus/dbus-message.c.orig	2017-01-26 20:32:02.456180239 +0900
 +++ dbus-1.10.10/dbus/dbus-message.c	2017-01-26 20:32:27.288180875 +0900
 @@ -4128,7 +4128,7 @@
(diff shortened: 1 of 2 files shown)


frznlogic:
That patch have pissed me off quite a bit... I'd love if they can be removed :)

gunnarx:
The patch messes with a few other things than readdir, but those changes seem not that useful.  

I fixed the spelling error in my message, but I won't push a cleanup until the first build results comes back.

gunnarx:
Hmm... PRs  not build-tested against x86 32-bit.  Added a pipeline, just to get this tested now.

frznlogic:
You have the same build issues I ran into locally. There is something with the optimization flags vs some internal flags. I didn't figure out how to solve it as I didn't have time :(

gunnarx:
Yes but it seems to have nothing do with the patch, so it should exist also on master (for 32 bit build)?  Anyway, I'll push a blacklist modification just to confirm if the error goes away.

gunnarx:
Latest failure: dbus-message.c:4372:36: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]

And of course I had to eat my words.  Looks like one part of the patch was actually necessary...  Apparently the sign mismatch problem is not triggered in a 64 bit build.  Patch needs to be refreshed if we keep it.

Oh and I guess the "blacklisting" of the hardening flags worked...

gunnarx:
- Adjusted patch again to include the integer cast that was originally proposed.  I have not checked that deeply - just keeping what was there already.

- Then there was an unused variable also which was caught by -Werror to become a failure.  I temporarily disabled the treating of this as an error for dbus compilation.  I checked and in later D-Bus versions this has been fixed by moving the declaration inside the #ifdef where it is conditionally used.  So when dbus component is updated in layers later on, then -Werror could be enabled again for unused vars (but that is a fairly strict setting to have...)

Guess it's ready for review now.

meta-ivi/recipes-core-ivi/dbus/dbus/build_fix-ubunt1604-qemux86.patch
```diff
@@ -30,15 +9,3 @@
      {
```
gunnarx, line 9: This part of the patch is unchanged..., the rest is removed.

gunnarx:
I had a quick chat with frznlogic on this, so I think I can go ahead and merge it.  It should fix the build-error for master and other PRs so even if the solution might not be perfect, it's worthwhile to merge it now.

meta-ivi/recipes-core-ivi/dbus/dbus/build_fix-ubunt1604-qemux86.patch
```diff
@@ -30,15 +9,3 @@
...
        int *a = dbus_realloc(loader->unix_fds,
                              loader->max_message_unix_fds * sizeof(loader->unix_fds[0]));
---- dbus-1.10.10/dbus/dbus-shell.c.orig	2017-01-26 20:45:49.292201410 +0900
-+++ dbus-1.10.10/dbus/dbus-shell.c	2017-01-26 20:47:39.840204241 +0900
-@@ -419,7 +419,9 @@
-             {
-             case '\n':
-               if (!delimit_token (&current_token, &retval, error))
-+		{
```
gunnarx, line 39: Skipped this since it serves no purpose.  This is a  style change, that should either just be accepted as it is, or changed upstream, not here.

PR #83, opened 2018-09-04
linux-yocto: remove deprecated CONFIG_EXPERIMENTAL

sashko:
CONFIG_EXPERIMENTAL has been removed in Linux 3.9. Proof: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5a958db311c416345a61b7bddc62a5c6c7112c7f

Signed-off-by: Oleksandr Kravchuk <oleksandr.kravchuk@pelagicore.com>

PR #84, opened 2018-09-07
ivi-image: remove hard-coded IMAGE_ROOTFS_SIZE

mirzak:
The image rootfs size is determined by the following algorithm
(from Yocto ref manual):

    if (image-du * overhead) < rootfs-size:
        internal-rootfs-size = rootfs-size + xspace
    else:
        internal-rootfs-size = (image-du * overhead) + xspace

Currently rootfs-size is set to 8 MB which essentially means that it
will never be used as it is considered as a minimum. By removing it
from the image, the weak default value (65536) will be used which is
defined in bitbake.conf.

The primary reason of removing it is to be able to set a custom value in
e.g local.conf which is currently not possible.

Signed-off-by: Mirza Krak <mirza.krak@northern.tech>

### DIFF
--- a/meta-ivi/recipes-yocto-ivi/images/ivi-image.inc
+++ b/meta-ivi/recipes-yocto-ivi/images/ivi-image.inc
@@ -9,8 +9,6 @@ LICENSE = "MIT"
 
 inherit core-image buildhistory
 
-IMAGE_ROOTFS_SIZE = "8192"
-
 IMAGE_FEATURES += " package-management"
 
 # Create SD image symlink correctly


mirzak:
This is also an candidate for back-port/cherry-pick to stable branches.

gunnarx:
I'm helping out @frznlogic a few days.  
Merged to branches **master**, **14.x-rocko** and **14.x-sumo**.

slawr:
@gunnarx good to see you thought of the 14.x branches. I almost commented on #83 about that and mirroring to yoctoproject.org but decided it might be in progress and so I would just ask about it in the next BIT call. At the moment master, 14.x-rocko and 14.x-sumo are not quite aligned, although the differences are minor.

PR #86, opened 2018-09-10
Drop packagegroup for x11/xserver

gunnarx:
This looks like an unnecessary (perhaps a bit augmented) copy of:
poky/meta/recipes-graphics/.../packagegroup-core-x11-xserver.bb

The package group is not included in the standard meta-ivi image, and
I'm not aware of any projects based on meta-ivi that are not also
Wayland-based.  (If such projects exist they can reference the poky file
with a possible bbappend, or make contact with the meta-ivi maintainers)


gunnarx:
Ooops looks like I was not properly up to date... rebased and pushed again.

PR #87, opened 2018-09-12
MAINTAINERS: fix maintainer's email

sashko:
Signed-off-by: Oleksandr Kravchuk <dev@sashko.rv.ua>

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -20,13 +20,12 @@ When creating patches, please use something like:
     git format-patch -s --subject-prefix='meta-ivi][PATCH' origin
 
 When sending patches, please use something like:
-    git send-email --to=yong-il.joh@windriver.com \
+    git send-email --to=oandreasson@luxoft.com \
                    --cc=genivi-meta-ivi@lists.genivi.org <generated patch>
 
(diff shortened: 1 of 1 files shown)


mirzak:
Is this workflow even valid, that is using "git send-mail". I thought changes come in trough PR requests on github? :)

sashko:
@mirzak I guess some people prefer it the old school way :)

slawr:
Gunnar and I were discussing a related point on a different PR from Oleksandr. I think this content comes from when the primary host for meta-ivi was yoctoproject.org and WR provided the maintainer. Gunnar was saying a review of the meta-ivi submission policy might be useful. Although of course that should be done with the current maintainer :) Even if they are happy still to receive patches via the ML it should mention github. Personally I think it would be nice to have some commonality with GDP as well.

That review shouldn't stand in the way of this PR though. Oscar is more likely to respond than Tolkien :)

frznlogic:
I agree, this is not the way I've been working at least :). I much prefer PR's in github as a workflow to email based management.

frznlogic:
I'll merge this for now.

PR #88, opened 2018-09-18
README.md: correct meta-ivi release cadence

slawr:
The Branch policy incorrectly states that meta-ivi is released biannually.
Correct it to state there is a release twice a year.

Signed-off-by: Stephen Lawrence <stephen.lawrence@renesas.com>


frznlogic:
My bad, I actually thought it could mean twice per year as well, which it can according to some sources, but it's definitely ambiguous 
https://www.thefreedictionary.com/biannually

PR #89, opened 2018-09-26
Prepare build-script to support physical hw target(s) + test image fixes

gunnarx:


gunnarx:
I recognize that this PR right now has a few new commits that should go to master, and that are **not** really related to this issue.  The script fix was however needed to make things build here.  I will move commits around and sort this soon.

slawr:
Thanks for preparing these @gunnarx. I need to update meta-ivi-renesas for sumo next week and there is the Tech Summit of course, so I might not get to advance my side of this until after the Summit.

gunnarx:
Cleaned up the commits and made meta-ivi-test layer inclusion conditional so that it only applies on a test-image build.  Let's see if I got it right...

gunnarx:
The PR is still mixing R-Car support with test-image fixes.  I've forgotten by now how we ended up there but never mind.  Confirming both features separately:

1. Build checks show that normal image builds are unaffected and successful for current targets (qemu x86, 86_64, and arm64).

2. Test image also built fine [on separate pipeline here](https://go.genivi.org/go/tab/build/detail/yocto-baseline-testimage/16/Build/1/init_and_bitbake)  (qemux86_64).  _This pipeline link might be removed if you reference this far in the future._

3. Testing R-Car build for the branch on a [preliminary pipeline that was set up](https://go.genivi.org/go/tab/build/detail/yocto-baseline-14-r-car-m3-starter-kit/1/Build/2/init_and_bitbake) -- NOTE access to that link could be limited unless you are authenticated, since 
A) we need to confirm that it is producing artifacts OK, and 
B) those artifacts may include graphics drivers that have distribution limitations.

gunnarx:
I have done additional amendments. The script adds to bblayers.conf to include the needed bblayers and sublayers from meta-renesas, meta-ivi-renesas and meta-linaro.

gunnarx:
Latest build [here](https://go.genivi.org/go/pipelines/yocto-baseline-14-r-car-m3-starter-kit/8/Build/1). Note same access restrictions as noted above.  

I think the build support in the script seems basically done now.  There is some build failure on optee however.  Possibly some adjustment in **meta-genivi-dev** also needs to be ported over to the baseline adjustment/append layer (**meta-ivi-renesas**)?  Transfer to @slawr for investigation.

slawr:
Thanks Gunnar.

> There is some build failure on optee however. Possibly some adjustment in meta-genivi-dev also needs to be ported over to the baseline adjustment/append layer (meta-ivi-renesas)? Transfer to @slawr for investigation.

My educated guess would be that the meta-linaro optee BBMASK was missing from the build, but that should not be the case if the local.conf from GDP is used. Anyway I'll investigate when I return to the CI work.

scripts/ci-build.sh
```diff
@@ -478,21 +522,78 @@ MIRRORS_append = \"\\
...
+    target_local_conf_file=
+    ;;
+esac
+
+
+if [ -n "$target_local_conf_file" ] ; then
+  # Download and use hardware-specific local conf settings (reused from the GDP project)
+  cd "$BASEDIR/build/conf"
+  local_conf_url="$GDP_TEMPLATES_URL/$target_local_conf_file"
+  fetch "$target_local_conf_file" "$local_conf_url"
```
slawr, line 579: I wonder if this is the cause of your build failure. It seems the fetch will copy to build/conf. Line 581 appends "require r-car-m3-starter-kit.local.conf" to your local.conf. However if you actually look at r-car-m3-starter-kit.local.conf it [includes renesas-rcar-gen3.local.inc](https://github.com/GENIVI/genivi-dev-platform/blob/cc28fbfb130548f4cc9bf549844f2d5464a1bfef/gdp-src-build/conf/templates/r-car-m3-starter-kit.local.conf#L2) from the sub-directory template in GDP which is not present in meta-ivi. Basically a difference in the build environment between meta-ivi and GDP. You have no template sub-directory in meta-ivi build/conf.

So I think at run-time in your environment it may not find the sub-directory so the majority of the required local.conf additions are not made. As the file uses an include, rather than require no error is thrown. A side effect of the missing local.conf is that [meta-linaro optee is not masked](https://github.com/GENIVI/genivi-dev-platform/blob/cc28fbfb130548f4cc9bf549844f2d5464a1bfef/gdp-src-build/conf/templates/renesas-rcar-gen3.local.inc#L23) which results in the CI failure you see.

That's based on a reading of the source. I don't have access to the CI build tree to check conclusively.

If right you could just extend the build script to create a template sub-directory and copy the GDP files there. Then the environments would be the same and it should work.

slawr (commented):
I think a difference between GDP and how you setup meta-ivi may be the cause of your build failure

scripts/ci-build.sh
gunnarx: I will check today but I'm quite sure you found it.   Now that you mention it I seem to remember having a similar thought, but never finished the thought it seems.  I amended 
67fef7e with this proposal.  Let's see.

slawr: My knowledge of the optee mask meant I could zero in quite quickly. Lucky in one regard that was the build failure :)

I almost just hacked the script to see if it fixed it but thought I could know for sure even quicker if I could access the build. So left it as you have that power.

slawr: Oh I just remembered something else I noticed when I looked at the failed build last night. It was complaining about the layer compatibility not being set for meta-ivi-renesas. I addressed that so either it was just a timing thing (the build was before I pushed the change) or if SHAs are hard coded it (script?) needs an update.

slawr: I just went back to consider the meta-ivi-renesas observation. You have this on meta-ivi master which is now moved to Thud. R-Car Thud community YBSP is not public yet. So this should move to sumo.

gunnarx: OK let's try #99 instead

PR #90, opened 2018-09-27
README: update dependency layers to latest revisions (Rocko branch)

sashko:
Signed-off-by: Oleksandr Kravchuk <oleksandr.kravchuk@pelagicore.com>

PR #91, opened 2018-09-27
README: update dependency layers to latest revisions (Sumo branch)

sashko:

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,12 +100,12 @@ Layer Dependencies
 
 URI: git://git.yoctoproject.org/poky
 > branch:   sumo
-> revision: 96fbd39ba32362416c18d90bb7a81eb6a76912e0
+> revision: d240b885f26e9b05c8db0364ab2ace9796709aad
 
 URI: git://git.openembedded.org/meta-openembedded
 > layer:    meta-oe
 > branch:   sumo
-> revision: b0950aeff5b630256bb5e25ca15f4d59c115e7c1
+> revision: 2bb21ef27c4b0c9d52d30b3b2c5a0160fd02b966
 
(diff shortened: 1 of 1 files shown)


gunnarx:
Thanks for this one, and #90.  Just one thing - since the baseline is required to be compliant with the GENIVI specifications, we normally check what has been changed before updating the included layers.  It is just a simple sanity check to see that nothing becomes "non-compliant" due to the upgrade.  I think it is rarely a problem and I would not really expect to see API breaking changes for example, within one openembedded branch, but it's worth a quick check.

For the meta-ivi baseline it has not been done fully, but I would like to start doing like GDP does, which is to include a short log of all the changes in the layer between the two hashes.  We could create a new script to make it easier...  

For GDP Here is an example of what the [commit message](https://github.com/GENIVI/genivi-dev-platform/commit/2414260fb4e7132fd067092e583abd40c3d43269) with shortlog looks like.
And [here is the script](https://github.com/GENIVI/genivi-dev-platform/blob/master/scripts/git-submodule-bump) that creates it (but it only works on the GDP submodules, so a new one would be needed here).

gunnarx (changes requested):
Requesting a change to include shortlog as noted in the previous comment.

gunnarx:
Can #90 be closed because this one is newer?  They modify the same layers, right?  Which one do you think we should aim for here?

sashko:
@gunnarx #90 is for `rocko` and this one is for `sumo`.

gunnarx:
Ah yes, I forgot about that... :)

gunnarx (approved):

gunnarx:
Great, the short log looks good.  I could not see anything that stands out as being an issue for compliance either so let's merge it.

PR #79, opened 2018-08-26
Adapt to Yocto "Thud" 2.6

sashko:
Signed-off-by: Oleksandr Kravchuk <oleksandr.kravchuk@pelagicore.com>

PR #92, opened 2018-10-26
Persistence and lifecycle updates 

gunnarx:
Note that the name of the PAS update is just slightly speculative since the 1.0.10 tag does not yet exist.  
Therefore WIP.  

But the SRCREV exists on the master branch so this recipe update will need to be done anyway.  But I see little reason that it will not be tagged exactly this way...

The other components have already received new version/tags by me.


gunnarx:
Looks like we jumped the gun on the NHM README rename...  I'm fixing it, and updating this PR.  Let's try again.

gunnarx:
Looks good, integrated into #79

PR #93, opened 2018-10-31
Outgoing/thud test

gunnarx:
Testing to avoid weston issue for qemux86*


gunnarx:
Yup, this worked.  Looks like #79 is updated now.

PR #95, opened 2018-11-16
README.md: Update poky commit to thud-20.0.0

gunnarx:
This is using the tagged/release commit of poky - thud.

Note:
- poky:thud branch has additional commits since the tag - we can try updating that separately
- meta-gplv2 was already at the latest thud.
- meta-openembedded also already at the latest thud.  As of today meta-openembedded also has additional commits on master that are not on thud, but those are presumably for "next".

PR #96, opened 2018-11-16
README.md: Update poky to tip of thud

gunnarx:
A separate PR to test also the latest tip of thud.  #95 explains the rest

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ URI: git://git.yoctoproject.org/poky
 URI: git://git.openembedded.org/meta-openembedded
 > layer:    meta-oe
 > branch:   thud
-> revision: f1511d254632a34c1deb51f4bf8b8c21e7423f51
+> revision: 4cd3a39f22a2712bfa8fc657d09fe2c7765a4005
 
 URI: git://git.yoctoproject.org/meta-gplv2
 > branch:   thud
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Layer Dependencies
 
 URI: git://git.yoctoproject.org/poky
 > branch:   thud
-> revision: 84eecb017ef92ef36b4df730908828e54aeff85c
+> revision: 50f33d3bfebcbfb1538d932fb487cfd789872026
 
 URI: git://git.openembedded.org/meta-openembedded
 > layer:    meta-oe
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ URI: git://git.openembedded.org/meta-openembedded
 
 URI: git://git.yoctoproject.org/meta-gplv2
 > branch:   thud
-> revision: 379ea8dd144b06aeb459e9a82c792c84d8a5baf7
+> revision: aabc30f3bd03f97326fb8596910b94639fea7575
 
 Using the above git SHAs and the master meta-ivi branch,
  bitbaking pulsar-image is known to work


gunnarx:
Just note that I didn't put this on top of #95.  \***If**\* we take both of them, which might be good in case of some regression, then I let's have them in the right order with 20.0.0 tag first, then this.

gunnarx:
Yes also the tip of thud branch for poky has the problem we identified in #79. (for the qemu-x86 targets only).  Here is [the commit](https://git.yoctoproject.org/cgit/cgit.cgi/poky/commit/?id=4001e1) we were concerned about.  I guess we should start discussing that with the OE core devs to sort it out.

gunnarx:
On second thought, meta-ivi has a [bbappend on weston](https://github.com/GENIVI/meta-ivi/blob/13f502d06ef7732dc5f7c5de767141450496e4f6/meta-ivi/recipes-graphics/wayland/weston_%25.bbappend), which installs weston.ini and it therefore needs to be fixed there.  It's not a bug in OE-Core, just a consequence of this getting rearranged in the upstream project.  

If the custom weston.ini installation is moved to a weston-conf_%.bbappend instead, that should probably take care of it.

slawr:
> On second thought, meta-ivi has a [bbappend on weston](https://github.com/GENIVI/meta-ivi/blob/13f502d06ef7732dc5f7c5de767141450496e4f6/meta-ivi/recipes-graphics/wayland/weston_%25.bbappend), which installs weston.ini and it therefore needs to be fixed there. It's not a bug in OE-Core, just a consequence of this getting rearranged in the upstream project.
> 
> If the custom weston.ini installation is moved to a weston-conf_%.bbappend instead, that should probably take care of it.

As an aside this (adapting to weston recipe changes in YP poky meta) would also apply to GDP when you come to do Thud. There has been a long standing JIRA ticket related to GDP not following the poky recipe form for using weston-init for confs which causes issues when combined with other layers. Addressing that ticket on Thud would mean GDP should take account of weston-init and weston-conf. Ticket is [GDP-530](https://at.projects.genivi.org/jira/browse/GDP-530)

gunnarx:
> As an aside this (adapting to weston recipe changes in YP poky meta) would also apply to GDP when you come to do Thud

Yes.

>There has been a long standing JIRA ticket related to GDP not following the poky recipe form for using weston-init for confs which causes issues when combined with other layers. 

OK, thanks for the reference.  Following the same pattern would make sense, but that's another part isn't it?  The file that is conflicting here (weston.ini) appears to be handled by weston-conf recipe, not weston-init.  OK, I guess that's why you called it an aside :-)

>Addressing that ticket on Thud would mean GDP should take account of weston-init and weston-conf.

I think you are mostly planting a reminder for that other ticket.  I was about to go and write it there instead of here, but I see you did that already.

slawr:
>OK, thanks for the reference. Following the same pattern would make sense, but that's another part isn't it? The file that is conflicting here (weston.ini) appears to be handled by weston-conf recipe, not weston-init. OK, I guess that's why you called it an aside :-)

>I think you are mostly planting a reminder for that other ticket. I was about to go and write it there instead of here, but I see you did that already.

Yes, yes, yes and yes :)

After planting the reminder I went and found and updated the ticket. I wrote a reply, but I'll post it there instead.

gunnarx:
To summarize again:

failure :
```
file /etc/xdg/weston/weston.ini conflicts between attempted installs of weston-5.0.0-r3.core2_64 and weston-conf-1.0-r0.qemux86_64
```

It is the bbappend that meta-ivi introduces onto weston that causes this because this bbappend installs its own **weston.ini**, but since the thud changes in meta-oe it is now **weston-conf** that is responsible for installing weston.ini.  (Note special MACHINE overrides **do_install_qemux86**(), and same for x86-64,
in poky/meta/**weston-conf.bb** if you are interested.  Therefore only qemu-x86* targets fail).

**Try:**
In **meta-ivi**, try moving the weston.ini installation done now in **weston_%.bbappend**
to **weston-conf_%.bbappend** (new file) instead.  Maybe consider how it might interact with the MACHINE overrides, or just try it...

Finally, this PR could be updated before merging since "tip" of thud has likely moved on.

martin-ejdestig:
@gunnarx I pushed a fixup commit to #95 (continuing discussion here since more details) that adds a bbappend for weston-conf.bb.

When trying to build that I get an error though:
```
install: cannot stat '/mnt/storage/mejdestig/meta-ivi-master/build-test-qemux86-64/tmp/work/qemux86_64-poky-linux/weston-conf/1.0-r0/weston-conf-1.0/ivi-shell/weston.ini.in': No such file or directory
```

I do not see how ivi-shell/weston.ini.in ends up in work directory of weston, so I can not do the same for weston-conf. Do you know how this works?

martin-ejdestig:
> I do not see how ivi-shell/weston.ini.in ends up in work directory of weston, so I can not do the same for weston-conf. Do you know how this works?

To clarify a bit here, I have searched through all source directories and the build directory and am not able to see where the directory comes from.

martin-ejdestig:
OK, found it now. ivi-shell/weston.in.ini actually comes from the weston-5.0.0.tar.xz archive itself. (I did not know that ivi-shell was in Weston upstream.)

So, how do we want to handle this? We can not add a dependency to the weston recipe in the weston-conf recipe (I am guessing... circular dependency). Keep a copy of ivi-shell/weston.ini.in in meta-ivi next to our bbappend to weston-conf?

slawr:
~~As an aside I'm currently setting up my R-Car Thud builds so I can take a look as well. Wondering if its best to start with the current master tip or jump straight to the last releases of poky/meta-oe and one of these PRs. Which PR is the consolidation point for the Thud update?~~
Update: ok I thought the PRs had code changes as well. I can see why you have the two PRs - different SHA references. You're going to be jumping to something fresh and I think the issue below is more related to Qemu so I'll integrate to the latest poky/meta-oe releases.

> So, how do we want to handle this?

To be honest I'm a little lost as to exactly what 'this' is now :), what with the discussions and code spread over multiple PRs. 

Based on quick appraisal, I looked at PR #95 (as you ref it recently) and your last error:
https://github.com/GENIVI/meta-ivi/pull/96#issuecomment-470502016 on assumption that is latest code and error you are trying to fix?

Assuming your QEMU provides its own weston.ini then I think the error may be simply because you don't have it in the right place. It should be part of weston-conf, whilst you had it in weston, i.e. your fix-up didn't move the weston.ini as well.

See R-Car as an example of Thud weston-conf, weston-init, weston recipe when you have your own weston.ini:
https://github.com/renesas-rcar/meta-renesas/tree/thud-dev/meta-rcar-gen3/recipes-graphics/wayland

martin-ejdestig:
> Assuming your QEMU provides its own weston.ini then I think the error may be simply because you don't have it in the right place. It should be part of weston-conf, whilst you had it in weston, i.e. your fix-up didn't move the weston.ini as well.

ivi-shell/weston.ini.in "template file" is part of the weston tarball, so it is not available when bitbake:ing the weston-conf recipe. So if we want to continue to use that, we have to e.g. copy it.

I think we are using the template in a way it is not meant to be used though. Supposed to be part of the build of weston. Build system does replacement in it: https://gitlab.freedesktop.org/wayland/weston/blob/master/ivi-shell/meson.build#L47 . They have also changed the name of the template variables in a recent commit (something else that hints that we are using it in a way it is not supposed to be used, I think): https://gitlab.freedesktop.org/wayland/weston/commit/67546bed04a464d18508265976f2a4b673db565c

> See R-Car as an example of Thud weston-conf, weston-init, weston recipe when you have your own weston.ini:
> https://github.com/renesas-rcar/meta-renesas/tree/thud-dev/meta-rcar-gen3/recipes-graphics/waylan

Yes, so this is only a problem if you want to use https://gitlab.freedesktop.org/wayland/weston/blob/master/ivi-shell/weston.ini.in as a template.

If we want to do that for our QEMU builds, as was done before, it is still problematic even for QEMU.

martin-ejdestig:
Since #95 has been merged now I rebased this to master and updated hash for poky to current tip of thud (with --date=now as to not lure people into thinking @gunnarx was from the future ~5 moths ago :).

Also updated meta-openembedded and meta-gpl2 to what their thud branch points to as of today.

gunnarx:
4cacd9c4920ad7114587be712dc24a9e13d34e32 built OK
[Build log here](https://go.genivi.org/go/tab/build/detail/yocto-baseline-poky-thud_tip-test/2/Build/1/init_and_bitbake).  I set it up a temporary pipeline because the PR pipeline didn't build. So the link might go away eventually.

PR #97, opened 2018-11-23
bluez5: enable obex-profiles

martin-ejdestig:
Without this --disable-obex is passed to configure and no bluez5-obex*.rpm
is generated. packagegroup-abstract-component-p1.bb in
meta-ivi/recipes-yocto-ivi/packagegroups/ RDEPENDS on bluez5-obex.

Error without this patch when enabling meta-ivi-test and bitbake:ing
test-image:

Error:
 Problem: conflicting requests
  - nothing provides bluez5-obex needed by packagegroup-abstract-component-p1-1.0-r0.noarch

### DIFF
--- a/meta-ivi/recipes-connectivity/bluez5/bluez5_%.bbappend
+++ b/meta-ivi/recipes-connectivity/bluez5/bluez5_%.bbappend
@@ -1,5 +1,7 @@
 FILESEXTRAPATHS_append := ":${THISDIR}/${PN}"
 
+PACKAGECONFIG_append = " obex-profiles"
+
 #
 # Service API reference implementation
 # Revision: 5.15 (Horizon)


martin-ejdestig:
Unsure how this has ever worked. The obex-profiles feature has been like this for a long time in poky (since a767e362580b68dfc50fb3c54289f97558abb60c more than 5 years back it looks like to me) and I do not see it ever having been enabled in meta-ivi. What am I missing?

gunnarx:
> What am I missing?

The obex profiles, it looks like :) :)

I can't really understand it either on a quick look.  Let's see if we can sort this out next time we talk.  Looks safe enough as a change though.

gunnarx (approved):

PR #98, opened 2018-11-27
ecryptfs-utils: remove recipe and drop the package

sashko:
Removed recipe from the layer as it is already maintained by the
meta-security.

(cherry picked from commit 86209a75e66580f46eb7261522e36f61003d2e03)


slawr (commented):
Hi Oleksandr,

I can see the usefulness of not maintaining a recipe in meta-ivi where one is maintained upstream. Particularly when it is not a Genivi maintained component.

That said your PR mentions that it is available in meta-security, which is currently not listed as a layer meta-ivi is dependent on. So that has consequences. On a simple level it would mean the meta-ivi documentation needs to be updated to include it. Trivial, but a task. On a more complex level understanding what the impact of using meta-security with the baseline is, both that there are no layer integration issues due to its recipes and they are configured appropriately for the baseline.

My working assumption is that you've been using meta-security with the baseline. Perhaps in the context of Pelux? Any input from your side on your findings? Package configurations perhaps?

gunnarx:
@slawr  We actually already accepted this on master, see https://github.com/GENIVI/meta-ivi/commit/86209a75e66580f46eb7261522e36f61003d2e03 -- (the commit makes a small note of this being a cherry pick).

When it was proposed last time I did not find any requirement suggesting that it should be part of the -p1-specific group, and we found nothing else that depended on this component.  So, therefore it was removed from the -p1 group.   When that is done there should be no need to depend on the meta-security layer.

So the question of this PR is rather: Which things should be backported to a maintenance branch?  I advocate a policy that is roughly speaking to backport only that which is "important" : bugfixes or obvious flaws.   This seemed more like a small quality improvement, and I therefore decided (in my interim time as maintainer) to keep it only on master until the next release when it becomes part of the mainstream.

However, now I assume this is actually conflicting in builds because of meta-security layer? 
Can you clarify @sashko if this is conflicting?  If it is then we have at least one that uses the sumo maintenance branch that has a problem here, and that sounds like an argument to accept the commit on the maintenance branch, IMHO.

gunnarx:
I restarted the builds after a small fix - seems the pipeline had a bug that turns up occasionally when a log file is not being created.   Should work now I hope.

slawr:
> @slawr We actually already accepted this on master, see [86209a7](https://github.com/GENIVI/meta-ivi/commit/86209a75e66580f46eb7261522e36f61003d2e03) -- (the commit makes a small note of this being a cherry pick).
> 
> When it was proposed last time I did not find any requirement suggesting that it should be part of the -p1-specific group, and we found nothing else that depended on this component. So, therefore it was removed from the -p1 group. When that is done there should be no need to depend on the meta-security layer.

Oh ok. If the removal does not mean meta-security must be added to the required layers then my comments are void. No need to answer my questions @sashko. I was extrapolating from the limited info in the PR text.

gunnarx:
@martin-ejdestig You were going to discuss with @sashko if this patch is needed in sumo branch or if it's enough to keep it on master (as it already is) for future releases?

martin-ejdestig:
Yes, @sashko is still on vacation though. I think he is back next week, will check with him then.

PR #99, opened 2018-12-04
Prepare for R-Car community builds of baseline

gunnarx:
Please see #89 for details.  
This PR for sumo instead of master, and only the relevant commit (I think that one is enough without the other script fixes, we'll see)

PR #102, opened 2019-02-06
#99 for master

martin-ejdestig:
#99 for master.

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -87,6 +87,25 @@ EOT
   fi
 }
 
+# Same principle as described above but for bblayers
+append_bblayers_conf() {
+  BBLAYERS_CONF="$BASEDIR/build/conf/bblayers.conf"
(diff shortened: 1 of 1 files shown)


martin-ejdestig:
@slawr @gunnarx Do you think it is worth waiting for CI/CD for this or should I just merge it?

slawr:
I think it should be safe. These are solid improvements to the CI build scripting from Gunnar and they would be useful on Master as well. 

An update will be needed when Thud support for R-Car is released (very soon) but I don't see that as a reason to hold back, but you could wait for that if you wished.

gunnarx:
My reasoning would be that it needs build & test at some point.  

I don't think the merge to master is a major risk, and in the very unlikely event that it's buggy, it can be fixed on master **but** that only happens if master branch is built immediately after merge to notice  the problem... which also requires CI.  

So I would say let's either confirm manually or wait for CI to be up.

gunnarx:
Hmm, guess I commented too late, I didn't realize.

PR #100, opened 2019-01-23
meta-ivi-test and runqemu fixes

martin-ejdestig:

PR #103, opened 2019-02-20
meta-ivi-test: Suggest slirp for non-root qemu usage

gunnarx:
Based on today's discussion


gunnarx:
Is the suggested command correct @martin-ejdestig

martin-ejdestig:
Yes, looks correct.

But, could you perhaps put this commit on #100 instead? Small conflict but instructions on master is wrong (-i). (And maybe have a quick look at my commits, and, if you think it looks OK, I'll merge all three of them at the same time asap.)

martin-ejdestig:
Went ahead and cherry picked to #100.

PR #104, opened 2019-03-11
dlt-daemon: bump to 2.18.1

sashko:
Changed SRC_URI to release tarballs as it is faster and more intuitive.

Insallation of dlt-dbus.service was removed since package is built
with WITH_DLT_DBUS=OFF by default now (see @1b3c8b3d10fbc).

Also removed mainstreamed 0001-Fix-compilation-with-glibc-2.28.patch.


meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.18.1.bb
```diff
@@ -12,26 +12,25 @@ SECTION = "console/utils"
...
 inherit autotools gettext cmake systemd
 
 PACKAGES += "${PN}-systemd"
 SYSTEMD_PACKAGES = "${PN} ${PN}-systemd"
 SYSTEMD_SERVICE_${PN} = "dlt-system.service dlt.service"
 SYSTEMD_AUTO_ENABLE_${PN} = "enable"
-SYSTEMD_SERVICE_${PN}-systemd = "dlt-example-user.service \
-    dlt-dbus.service \
+SYSTEMD_SERVICE_${PN}-systemd = " \
+    dlt-example-user.service \
```
martin-ejdestig, line 33: So this means the DLT systemd service will be called dlt-example-user.service in a running system? We do not want that?

martin-ejdestig (commented):
> Insallation of dlt-dbus.service was removed since package is built
with WITH_DLT_DBUS=OFF by default now (see @1b3c8b3d10fbc).

Do you mean WITH_DLT_SYSTEMD here? Also, we want to build with WITH_DLT_SYSTEMD set to ON so that we get a service?

Also, what does WITH_DLT_DBUS=OFF/ON actually mean?

gunnarx:
@ssugiura (new DLT-daemon maintainer) and @clipka (old DLT-daemon maintainer) — please be aware of this PR, and please comment if you have an opinion, and/or if you can help guide a decision.

meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.18.1.bb
```diff
@@ -12,26 +12,25 @@ SECTION = "console/utils"
 LICENSE = "MPLv2"
 LIC_FILES_CHKSUM = "file://LICENSE;md5=8184208060df880fe3137b93eb88aeea"
 
-PR = "r2"
-
 DEPENDS = "dbus zlib gzip-native"
 
-SRCREV = "bdde2374efd567f27d7a4ec70911b34cda061d27"
-SRC_URI = "git://github.com/GENIVI/${BPN}.git;protocol=https \
+SRC_URI = "https://github.com/GENIVI/dlt-daemon/archive/v${PV}.tar.gz \
```
gunnarx, line 17: I am OK with this.  It hurts from one perspective because I think it would make some sense to use git for all sources all the time.  But the way things are done in OE/yocto I have to admit that tarballs are more straight forward (git repos use weird storage tricks in the downloads cache, and unexpected checks to see if a tag has moved messes up BB_NO_NETWORK setups).  Tar is just easier to handle in OE/yocto world.

We have no particular documented policy for meta-ivi as far as I know.

meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.18.1.bb
```diff
@@ -12,26 +12,25 @@ SECTION = "console/utils"
...
     file://0002-Don-t-execute-processes-as-a-specific-user.patch \
     file://0004-Modify-systemd-config-directory.patch \
-    file://0001-Fix-compilation-with-glibc-2.28.patch \
-    "
-S = "${WORKDIR}/git"
+"
+SRC_URI[md5sum] = "6f2f6d3386991ed5d5e5078647c62ce5"
+SRC_URI[sha256sum] = "07f7a65d3d248e8f5eed820cae0e2ec81318417a81540647346f81cfe024ee65"
```
gunnarx, line 22: I personally think double checksums are a bad practice because it only causes more maintenance effort.  Again, we don’t have a decided policy however.  (I’m not sure if OE/Yocto ecosystem has agreed on a best practice, but I hope it is a single checksum?)

meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.18.1.bb
martin-ejdestig: Nm, I see now that this is only for the example and dlt.service is still added above. This is only example services?

sashko: seems to be: https://github.com/GENIVI/dlt-daemon/blob/master/systemd/dlt-example-user.service.cmake

meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.18.1.bb
sashko: Virtually every recipe I have seen has had both md5 and sha256 checksum set, so this seems to be a common practice.

martin-ejdestig:
Test images fail to build due to conflict between `meta-ivi-test/recipes-extended/dlt-daemon/dlt-daemon/0001-change-file-position-for-dlt-test-filetransfer.patch` and https://github.com/GENIVI/dlt-daemon/commit/dca8ab254aa0a687c32009079d85e4d8f960b213 .

martin-ejdestig:
@sashko I put a squash commit (so it is easy for you to see what I have done) on your branch that fixes the dlt-daemon tests. Please squash and push.

Ready to merge otherwise, I think.

meta-ivi/recipes-extended/dlt-daemon/dlt-daemon_2.18.1.bb
martin-ejdestig: Yes, very common. Unclear to me why though. Could not find a reason in https://www.yoctoproject.org/docs/latest/mega-manual/mega-manual.html .

martin-ejdestig: > I’m not sure if OE/Yocto ecosystem has agreed on a best practice, but I hope it is a single checksum?

Looking only at poky (@9dfebdaf7a):
```bash
$ grep -rI SRC_URI.*md5sum\]\ =\ \" . | wc -l
498
$ grep -rI SRC_URI.*sha256sum\]\ =\ \" . | wc -l
498
$ grep -rI -A 1 SRC_URI.*md5sum\]\ =\ \" . | grep sha256sum | wc -l
496
```

Seems like they think best practice is both.

PR #106, opened 2019-03-14
MAINTAINERS: remove maintainer's email

sashko:
Maintainers tend to change, but their names and emails tend to stay in the
contributing instructions for much longer.

Also, this makes no sense to CC a person who is actively using the designated mailing list, so remove maintainer's email address and
leave mailing list only.


sashko:
@martin-ejdestig or should we remove one's email completely and use mailing list only?

martin-ejdestig (approved):

PR #107, opened 2019-03-29
vehicle-signal-interface: add recipe

sashko:

### DIFF
--- a/meta-ivi/recipes-core-ivi/vehicle-signal-interface/vehicle-signal-interface_git.bb
+++ b/meta-ivi/recipes-core-ivi/vehicle-signal-interface/vehicle-signal-interface_git.bb
@@ -1,0 +1,35 @@
+SUMMARY = "A new software bus for rapidly accessing vehicle data"
+DESCRIPTION = "VSI allows very flexible high speed communication of \
+data between threads or tasks on a single processor node system."
+HOMEPAGE = "https://github.com/GENIVI/vehicle_signal_interface"
+SECTION = "libs"
+LICENSE = "MPLv2"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=9741c346eef56131163e13b9db1241b3"
+
+FILESEXTRAPATHS_append := ":${THISDIR}/${PN}"
+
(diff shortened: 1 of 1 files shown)


martin-ejdestig:
@sashko This PR was discussed a bit on today's BIT meeting. We want it, but just curious why you need it.

Also, it is not built when building the image now. Could you add a packagegroup-vehicle-signaling, add vehicle-signal-interface to it and then add the group to the image?

sashko:
> @sashko This PR was discussed a bit on today's BIT meeting. We want it, but just curious why you need it.
> 
> Also, it is not built when building the image now. Could you add a packagegroup-vehicle-signaling, add vehicle-signal-interface to it and then add the group to the image?

We want it because we would like to write some software that uses it. Since it's GENIVI project, I thought it would make sense to add the recipe to the layer that is specifically made for GENIVI components.

Sure, I'll create the packagegroup and add a new commit to the PR, but are you sure you want it as part of the image by default even though nothing is currently using it?

slawr:
Thank you for your continued support of the project Oleksandr.

As an aside we should also consider what it means to have VSI in meta-ivi at runtime. Should it be started by default? Does that make sense, or is it something that is better specified in a wider distro that consumes meta-ivi? If yes how and with what setup (in a way that allows it to be modified)? 

I don't think that is anything that would be a blocker for the PR to be merged, but it is something that should be considered. Possibly as a next step.

gunnarx:
Discussed today. Via @martin-ejdestig we received info that this is not immediately urgent to merge.  I'd rather merge the patch in the vehicle_signal_interface than carry the patch in the layer, so I'll try to do that and update the PR.

slawr:
> As an aside we should also consider what it means to have VSI in meta-ivi at runtime. Should it be started by default? Does that make sense, or is it something that is better specified in a wider distro that consumes meta-ivi? If yes how and with what setup (in a way that allows it to be modified)?
> 
> I don't think that is anything that would be a blocker for the PR to be merged, but it is something that should be considered. Possibly as a next step.

Discussed in the BIT call today. VSI is a library, rather than a daemon. So no need for a default service to start it.

sashko:
@slawr @gunnarx ping.

PR #109, opened 2019-05-10
Rename image from pulsar-image to meta-ivi-image

martin-ejdestig:
There is no suitable code name for the 15.0.0 release. Use a more generic image name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,8 +112,7 @@ URI: git://git.yoctoproject.org/meta-gplv2
 > revision: aabc30f3bd03f97326fb8596910b94639fea7575
 
 Using the above git SHAs and the master meta-ivi branch,
- bitbaking pulsar-image is known to work
- (the pulsar-image build should be aligned with GENIVI 14.0).
+ bitbaking meta-ivi-image is known to work.
 
 For creating a specific GENIVI compliant image version, please make sure you
 git checkout the related meta-ivi branch and follow the build instructions
--- a/meta-ivi/recipes-yocto-ivi/images/meta-ivi-image.bb
+++ b/meta-ivi/recipes-yocto-ivi/images/meta-ivi-image.bb
@@ -1,5 +1,5 @@
 # Base this image on generic IVI image
 require recipes-yocto-ivi/images/ivi-image.bb
 
-PV = "14.0.0"
+PV = "15.0.0"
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -112,8 +112,7 @@ URI: git://git.yoctoproject.org/meta-gplv2
 > revision: aabc30f3bd03f97326fb8596910b94639fea7575
 
 Using the above git SHAs and the master meta-ivi branch,
- bitbaking pulsar-image is known to work
- (the pulsar-image build should be aligned with GENIVI 14.0).
```
martin-ejdestig, line 116: OK with removing this?

martin-ejdestig: (Mention of a specific release that is.)

slawr: See no issue with that.

meta-ivi/recipes-yocto-ivi/images/meta-ivi-image.bb
```diff
@@ -1,5 +1,5 @@
 # Base this image on generic IVI image
 require recipes-yocto-ivi/images/ivi-image.bb
 
-PV = "14.0.0"
+PV = "15.0.0"
```
slawr, line 4: Personally I don't feel the rename is to bad. When you open this bb the comment inside says that ivi-image.bb is the generic part.

I would have to think/look more deeply but I believe this separation originally was the idea that ivi-image represented the 'core' that OSS or commerical offerings could consume and pull into their own definitions.

So the fact that we move from a code name suffixed baseline image name, e.g. pulsar-image, to an unchanging one with meta-ivi-image does not undermine that I think. I guess we could argue for additional comment to clarify the differences but that could be done later.

slawr (commented):
Frankly I unfortunately don't have time to check through the code to see if everything was caught. So it is not a full review. The changes here seem ok though.

With regards Gunnar's idea to move to renaming the image I approve.

martin-ejdestig:
@slawr @gunnarx I assume you are busy at the AMM and have not had time to look into this deeply enough to feel comfortable with approving.

I will go ahead and merge this, extract build info so I can put it in the Wiki and branch and tag etc. Everything looks OK to me. (Fixing errors in releases is what brown paper bags are for. :)

PR #111, opened 2019-07-22
persistent-client-library: add persistence-administrator to RDEPENDS

martin-ejdestig:
The persistence-client-library README contains the persistence-administrator daemon running as a precondition:

"Persistence Administration Service is available in the system and running"

### DIFF
--- a/meta-ivi/recipes-extended/persistence-client-library/persistence-client-library_1.2.0.bb
+++ b/meta-ivi/recipes-extended/persistence-client-library/persistence-client-library_1.2.0.bb
@@ -38,4 +38,7 @@ FILES_${PN}-dev += " \
     ${libdir}/libpersistence_client_library.so \
     "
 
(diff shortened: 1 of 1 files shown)


martin-ejdestig:
I see now that I might been a bit too hasty. It is only required for pcl tests, it looks like...?  I guess this patch does not make any sense then. Or?

Hum, does it ever make sense to install persistence-administrator on a production system?

gunnarx:
persistence-admin offers functions such as backing up and restoring the persistence store, clearing user personal data, ... and these are functions are intended for / required by, some production systems. I think pers-admin also performs some set up of the storage infrastructure, so I don't think (from what I remember...) that the framework is designed to work without it.

gunnarx:
For the way things are set up in the meta-ivi system, I think this is accurate.  Persistence administrator is required.  There might be other ways to rework this code in the future but I think this is likely to be accurate so I'll merge.

PR #112, opened 2019-07-22
Update PCO and PCL to new version (changed D-Bus installation policy)

gunnarx:
Updated versions (note, new patch level tag was created on each of the components)

Changes according to details given by Martin here:
https://github.com/GENIVI/persistence-common-object/pull/8

It is intentional to remove the do_install_append() here.  I should probably have mentioned in the commit comment that:  It is no longer needed to remove sysconfdir to avoid conflict.


gunnarx:
Rebased.

gunnarx:
@martin-ejdestig, does this look right to you?

...never mind, it should be correct

gunnarx:
There were conflicts still...?
OK, well it is properly rebased now.

martin-ejdestig:
Hashes and removal of `do_install_append()` looks correct to me.

PR #115, opened 2019-09-23
audiomanager: Remove custom do_qa_staging

gunnarx:
The build script had a custom definition of do_qa_staging
which now seems to be failing with this error:

   Exception: NameError: name 'qa_check_staged' is not defined

Something probably changed in OE/Poky that makes this code no longer
applicable.  If QA is not generally failing on a normal build then I
don't see a strong reason to keep this special code so the easiest
approach seems to remove it from the recipe.


gunnarx:
Seems to test OK -> merging...

PR #116, opened 2019-11-20
dlt-daemon: bump to 2.18.4 version

sebastienraillet:
The SRC_URI checks are update to match the 2.18.4 version.

Since 2.18.3, the installation of systemd services was removed when
the corresponding package isn't built (e.g dlt-system.service isn't
installed if the dlt-daemon isn't compiled with WITH_DLT_SYSTEM=ON).
See @514717f2.
Therefore, I use the PACKAGECONFIG facility of yocto to choose which
service file should be installed based on which feature has been
activated by the user.

As I use the PACKAGECONFIG facility, I also populate the CMake flags
based on which feature is activated or not.

Signed-off-by: Sebastien RAILLET <sebastien.raillet@marelli.com>


slawr:
Thanks for the PR Sebastien and the detailed commit msg which is very helpful. I'll bring it up in tomorrow's Genivi Baseline Integration Team meeting call.

ssugiura (approved):

gunnarx:
Great patch - thanks @sebastienraillet for upstreaming this.
Our build infrastructure needed some maintenance but I'm looking at that.  This PR should be build-tested soon and then we will merge it.

gunnarx:
Hi @sebastienraillet.  Sorry, it took some time to confirm this because there appeared to be another component failing on master in fact (if I'm not mistaken) and because the automated checks were not really working.  Anyway, I think your PR failed in the end, with the following message?

```
 Problem: conflicting requests
  - nothing provides dlt-daemon-systemd needed by packagegroup-specific-component-p1-1.0-r0.noarch
```
Should dlt-daemon-systemd also be removed from the packagegroup definition, or do you have some other idea of how to configure the component correctly to make this work?

I had to modify some persistence component, so that's why this is on a weird branch, but here is the full log if you are interested and the failure now seems to be because of this PR: 
https://go.genivi.org/go/tab/build/detail/yocto-baseline-master-r-car-m3-starter-kit-with-lava-test/31/Build/2/init_and_bitbake

sebastienraillet:
Hi @gunnarx,
Ah bad luck... 😞 I tried the link you give me but it seems I've no access. How can I have an access ?
Or at least, something which can allow me to redo the image generation on my side. 
Thanks in advance for your answer.

sebastienraillet:
Hi,
I made some modification which allow to fix the issue you had. Please, can you redo a test on your side. 
Thanks,

gunnarx:
Yes, this seems to work well now (when patching the unrelated issue on master).  Thanks @sebastienraillet for improving the dlt recipe in meta-ivi.

PR #117, opened 2020-01-20
common-api: Upgrade to latest versions available

sebastienraillet:
The common-api recipes (native and target ones) are update to the latest
upstream version available. SRC_URI checks are update to mach these new
versions.

The flag SANITY_REQUIRED_UTILITIES on java is moved inside the right
location. This flag is only working inside a distro.conf or local.conf
file. Now, you will really have an error if you try to launch a build of
the ivi image without java on your machine.

Signed-off-by: Sebastien RAILLET <sebastien.raillet@marelli.com>

### DIFF
--- a/meta-ivi/recipes-extended/common-api/capicxx-core-native_3.1.12.4.bb
+++ b/meta-ivi/recipes-extended/common-api/capicxx-core-native_3.1.12.4.bb
@@ -6,4 +6,5 @@ LAUNCHER_LINK = "capicxx-core-gen"
 require capicxx-native.inc
 
 SRC_URI = "https://github.com/GENIVI/${_BPN}-tools/releases/download/${PV}/commonapi-generator.zip"
-SRC_URI[sha256sum] = "979abe3c2ff29e6c5e8a61cdfbd262bf837e6a26201593c5e6e38160978f2a21"
+SRC_URI[sha256sum] = "6de1720a80c6b61c1f20512e03f86c29b21f4df572c89e1f9b2e38923177b3ea"
+
--- a/meta-ivi/recipes-extended/common-api/capicxx-dbus-native_3.1.12.2.bb
+++ b/meta-ivi/recipes-extended/common-api/capicxx-dbus-native_3.1.12.2.bb
@@ -6,4 +6,4 @@ LAUNCHER_LINK = "capicxx-dbus-gen"
 require capicxx-native.inc
 
 SRC_URI = "https://github.com/GENIVI/${_BPN}-tools/releases/download/${PV}/commonapi_dbus_generator.zip"
-SRC_URI[sha256sum] = "6fb53b10ae49952383c8eb990995c6a3197055ccb51b199c8570b89b73522669"
+SRC_URI[sha256sum] = "ee7c4bcffc38b9526ae179a057d39d3a2170622efddc924fc91cc42835d082f5"
--- a/meta-ivi/recipes-extended/common-api/capicxx-native.inc
+++ b/meta-ivi/recipes-extended/common-api/capicxx-native.inc
@@ -19,7 +19,6 @@ def get_launcher_name(d):
 LAUNCHER = "${@get_launcher_name(d)}"
 
 inherit native
-SANITY_REQUIRED_UTILITIES += "java"
 
 S = "${WORKDIR}"
 DD = "${D}${datadir}/${PN}-${PV}"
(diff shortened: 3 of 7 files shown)


meta-ivi/conf/distro/poky-ivi-systemd.conf
```diff
@@ -28,6 +28,8 @@ VIRTUAL-RUNTIME_initscripts = ""
 
 DISTRO_FEATURES_BACKFILL_CONSIDERED = "sysvinit"
 
+SANITY_REQUIRED_UTILITIES += "java"
```
Spinus1: Is it possible to specify a Java version? Some legacy tool are not working fine with new JRE...

sebastienraillet: Hello, 
I checked but sadly, the **sanity.bbclass** doesn't allow this :disappointed:

meta-ivi/recipes-extended/common-api/capicxx-core-native_3.1.12.4.bb
```diff
@@ -6,4 +6,4 @@ LAUNCHER_LINK = "capicxx-core-gen"
 require capicxx-native.inc
 
 SRC_URI = "https://github.com/GENIVI/${_BPN}-tools/releases/download/${PV}/commonapi-generator.zip"
-SRC_URI[sha256sum] = "979abe3c2ff29e6c5e8a61cdfbd262bf837e6a26201593c5e6e38160978f2a21"
+SRC_URI[sha256sum] = "6de1720a80c6b61c1f20512e03f86c29b21f4df572c89e1f9b2e38923177b3ea"
```
gunnarx: This seems to indicate a missing newline.

meta-ivi/conf/distro/poky-ivi-systemd.conf
gunnarx: @sebastienraillet What is the reason for moving this line from the capicxx-native.inc file to the distro definition here?

sebastienraillet: Hi @gunnarx,
Because if you check [the code](http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/sanity.bbclass#n655) of sanity.bbclass, you will see that this variable is a top level variable (like TMPDIR for example).

I also made a test by removing this line on a machine without java and yocto doesn't complain about a missing java. As soon as you add it inside the distro file, yocto will complain that java is missing, as expected.

meta-ivi/recipes-extended/common-api/capicxx-core-native_3.1.12.4.bb
sebastienraillet: Yes you're right. My fault I will fix it

sebastienraillet:
Hi @gunnarx, do you still need some clarifications from my side for this PR ? 
Thanks

gunnarx:
So, to recap:

- The previous location of SANITY_REQUIRED_UTILITIES probably had no effect.
- In the new location it will require java to be installed natively on the build host, or the build will fail early.
- But it seems to me that the standard builds (like in CI) seem to succeed fine without java installed(!)  So I am not convinced if it is right to include this flag or not.  I can only guess that they succeed because other components are such configured that they actually do not trigger the usage of the code generator. (Even if the **component recipe** says that it depends on it).

While we discuss if the flag should be included (in a place where it has effect), I would regardless of the outcome ask you to split the version upgrade and the moving of the flag into different commits.  It makes it easier to only take one part of the PR, or to revert if we have trouble later, and so on.

gunnarx:
> But it seems to me that the standard builds (like in CI) seem to succeed fine without java installed(!)

^ I just confirmed this on the CI system.  This PR as it is written, failed the build when the sanity check noted that java is not installed.  When I removed only that part of your change, then the build succeeds (same build machine, i.e. without java installed).

gunnarx:
@sebastienraillet Could you split the commit as requested? Or should I just do it as part of the merge?  I'm leaning towards skipping the java requirement line since it would fail our builds where it actually does not fail today - it requires java to be installed in the build infrastructure which we do not have today.  So I would not do it until the problem actually arises.   It's not ideal, I know...

sebastienraillet:
@gunnarx, I pushed again my modifications by removing the part about the java requirement.
I can understand your point of view about not breaking current build infrastructure. 
Do you want the creation of an another PR about the java requirement or we forgot this proposal ?

gunnarx:
That would be the best so that we have it stored if this question comes up again.  It would 
be enough to keep the commit here and I could drop it during merge, but yes the clearest should be another PR.  Much appreciated.

sebastienraillet:
Hi @gunnarx,
I've created the PR #119 to propose only the part regarded the SANITY_REQUIRED_UTILITIES for java.
Therefore, I consider this one as clean now on my side. Is it ok for you ?

gunnarx:
Perfect, thank you.

PR #119, opened 2020-02-21
Adds SANITY_REQUIRED_UTILITIES for java

sebastienraillet:
The flag SANITY_REQUIRED_UTILITIES on java is effective only
inside a distro.conf or local.conf file. Knowing that common-api
generators need java to work properly, you will now have an error
if you try to launch a build of the ivi image without java on your machine

### DIFF
--- a/meta-ivi/conf/distro/poky-ivi-systemd.conf
+++ b/meta-ivi/conf/distro/poky-ivi-systemd.conf
@@ -28,6 +28,8 @@ VIRTUAL-RUNTIME_initscripts = ""
 
 DISTRO_FEATURES_BACKFILL_CONSIDERED = "sysvinit"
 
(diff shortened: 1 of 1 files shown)


gunnarx:
First thanks for this. We had the conversation about this on mailing list a long time ago, but I should document the outcome here before closing.

- The standard CommonAPI code-generators indeed need a java runtime.
However:
- The minimal meta-ivi build does not appear to exercize them currently because it is possible to build the minimal system without java installed.    I believe the affected components that _do_ support common API are configured in a way that they not use it)

I therefore don't think it is right to add this line here - instead it might be added in the distro that makes use of meta-ivi.  Adding it would add an unnecessary requirement for everyone to install java for building the minimal image, which is not really required now.

PR #120, opened 2021-08-23
meta-ivi: assert hardknott LAYERSERIES_COMPAT

amstewart:
Assert that meta-ivi is compatible with OE-core/hardnott to support new product development on the latest OE stable branch.


gunnarx:
Please see #121 for explanation.

PR #121, opened 2021-08-23
cairo: make directory removal more resilient

amstewart:
Upstream OE-core commit 9bb2268677ac8f0c97433bf1f04555abe88028a9
implements a more intelligent removal system for several install
directories which the cairo bbappend in meta-ivi then tries to modify.

Because these directories may or may not contain files, depending on the
PACKAGECONFIG state, implement the same intelligent removal system from
OE-core, after the bbappend finishes its modifications.

Signed-off-by: Alex Stewart <alex.stewart@ni.com>

----

See also [ni/meta-ivi #2](https://github.com/ni/meta-ivi/pull/2).

## Testing
`cairo` and `cairo-native` now build in hardknott without issue.

### DIFF
--- a/meta-ivi/recipes-graphics/cairo/cairo_%.bbappend
+++ b/meta-ivi/recipes-graphics/cairo/cairo_%.bbappend
@@ -10,9 +10,10 @@ do_install_append () {
 	rm -f ${D}${bindir}/cairo-trace
 	rm -f ${D}${libdir}/cairo/libcairo-trace.so*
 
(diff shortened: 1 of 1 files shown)


gunnarx:
Thanks so much for providing this PR.  I apologize that it won't be merged, when you have done the work, and it looks very good.

Unfortunately this repository is now going to be archived and no longer maintained at this location.  (In other words, meta-ivi would need to be forked to a new location if further development is desired).

